PR #5, opened 2021-08-25
Supports multiple ElasticSearch/OpenSearch

fandigunawan:
Supports multiple ElasticSearch/OpenSearch connection string. There is a case where OpenSearch contains more than one node, therefore it is necessary to support multiple OpenSearch nodes.
E.g. SHUFFLE_OPENSEARCH_URL=http://os1:9200,http://os2:9200,http://os3:9200


frikky:
Hey! Sorry - this got totally forgotten about in the middle of everything. 

Your fix does indeed seem like it may work. We would really appreciate any other suggestions you may have for opensearch if you have a clustered setup. Thanks!

PR #7, opened 2021-09-05
Adds prefix in ES index

fandigunawan:
Adds prefix in ES index so that each Shuffle intances might utilize single instance of ES E.g instance1_workflowapp.

Set environment variable SHUFFLE_OPENSEARCH_INDEX_PREFIX to a value or otherwise it will create a default index name.

Screenshot from created indices with SHUFFLE_OPENSEARCH_INDEX_PREFIX=instance1

image

Screenshot from created indices if SHUFFLE_OPENSEARCH_INDEX_PREFIX is empty/not set
image

### DIFF
--- a/db-connector.go
+++ b/db-connector.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


frikky:
> Adds prefix in ES index so that each Shuffle intances might utilize single instance of ES E.g instance1_workflowapp.
> 
> Set environment variable SHUFFLE_OPENSEARCH_INDEX_PREFIX to a value or otherwise it will create a default index name.
> 
> Screenshot from created indices with SHUFFLE_OPENSEARCH_INDEX_PREFIX=instance1
> 
> image
> 
> Screenshot from created indices if SHUFFLE_OPENSEARCH_INDEX_PREFIX is empty/not set
> image

This is neat! I was thinking about it some time, but ended up not going for it, since the potential for breaking changes was too high. Will read through, test and merge :)

frikky:
This seemed good and working in my setup and will be part of the next nightly release. I've further added it to the default .env in order to highlight that it's possible :)

PR #11, opened 2022-11-17
Refactor proposition

davidhfive9s:
Here we have a small refactor of the shared.go file.
We have merged the two lines (resp.WriteHeader and resp.Write) into one function call (setResponse)
setResponse function is in a new file called utils.go
This function takes the response, status code and the message that need to be sent to the user 
We have only done it on the shared.go but it could be done on the other .go files
It reduces the length of the file from 16027 to 15489
We are also working on other point to refactor in order to reduce the length even more
Let me know what you think about this refactor and if we should do it on the other files

### DIFF
--- a/db-connector.go
+++ b/db-connector.go
@@ -2122,6 +2122,9 @@ func GetOrg(ctx context.Context, id string) (*Org, error) {
 					curOrg.Users = users
 					setOrg = true
 				}
+			} else if strings.Contains(err.Error(), `cannot load field`) {
+				log.Printf("[WARNING] Error in org loading, but returning without warning: %s", err)
+				err = nil
 			} else {
 				return &Org{}, err
 			}
--- a/shared.go
+++ b/shared.go
@@ -8340,7 +8340,7 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
 		log.Printf("[DEBUG] MFA login for user %s (%s)!", userdata.Username, userdata.Id)
 	}
 
-	tutorialsFinished := userdata.PersonalInfo.Tutorials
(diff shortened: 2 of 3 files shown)


frikky:
Hey David, and thanks for the pull request!

I kind of like what you're trying to do here, with simplifying two things into one. I'm personally against this change however, as it just adds unnecessary abstraction, making it more for new developers to learn that is "Shuffle specific" vs what the Go libraries are built for. 

As for file sizes and such: I've always loved large files vs. tons of small ones (opposite of most people :crying_cat_face: ) . You're absolutely right that Shared.go needs refactoring, e.g. by API type (apps, workflows, creators, etc), but not when abstraction is the only reason.

davidhfive9s:
Thanks for taking the time to respond !
I understand the fact that this abstraction is unnecessary I don't want to make you feel lost in your own code --'
What do you mean by refactoring by API type ? Reorganize in the file itself or create multiple new files (I'm guessing that is not it since you like large files) or even something else.

frikky:
> Thanks for taking the time to respond ! I understand the fact that this abstraction is unnecessary I don't want to make you feel lost in your own code --' What do you mean by refactoring by API type ? Reorganize in the file itself or create multiple new files (I'm guessing that is not it since you like large files) or even something else.

I do indeed mean to break it down! Right now a complaint internally has been to find the right thing as "shared.go" isn't really descriptive at all, hence makes it hard to intuitively find code. 
Example of what I was thinking for the next refactoring is that shared.go becomes:
- workflows.go
- apps.go
- users.go
- orgs.go
- docs.go
- environments.go

Probably a few more as well

PR #33, opened 2023-09-04
fix: removing unnecessary images

0x0elliot:


0x0elliot:
do a deeper clean up before merge

PR #34, opened 2023-09-07
feat: Encryption with cache

0x0elliot:
in works


0x0elliot:
don't merge. give me sometime.

0x0elliot:
thumbs up, but size might be an issue? also we need tests now that we have introduced this.

PR #36, opened 2023-10-04
feat: adding basic stats feature

0x0elliot:


db-connector.go
```diff
@@ -3844,6 +3844,114 @@ func GetUser(ctx context.Context, username string) (*User, error) {
...
+
+	if project.DbType == "opensearch" {
+		err = indexEs(ctx, nameKey, fmt.Sprintf("%d", stat.Timestamp), data)
+		if err != nil {
+			return err
+		} else {
+			log.Printf("[INFO] Successfully updated ops dashboard stats")
+		}
+	} else {
+		k := datastore.NameKey(nameKey, fmt.Sprintf("%d", stat.Timestamp), nil)
```
frikky, line 3944: So we save each item and dump them out based on timestamp? I think this could work, nice!

PR #46, opened 2023-11-03
Fix file write mode in uploadFile function

naveen521kk:
Previously this function used append mode to open the file being updated which caused issue of duplication of data in the file. This commit changes the mode to open file to write mode as well as enable truncation of the file.

Fixes https://github.com/Shuffle/Shuffle/issues/1176

I tested it locally, please take a look at the video where the above issue doesn't happen.

https://github.com/Shuffle/shuffle-shared/assets/49693820/6dce185f-ab4c-4f13-9865-7c90aede7ef3

### DIFF
--- a/files.go
+++ b/files.go
@@ -1173,7 +1173,7 @@ func uploadFile(ctx context.Context, file *File, encryptionKey string, contents 
 		} else if file.StorageArea == "s3" {
 			log.Printf("SHOULD UPLOAD TO S3!")
 		} else {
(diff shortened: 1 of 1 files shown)


Hithesh2022:
@naveen521kk  did you click on refresh button and saw like if u didnt click on refresh button its work fine but click on refresh and see the file

Hithesh2022:
I too came through github octernship and faced same issue

naveen521kk:
> @naveen521kk did you click on refresh button and saw like if u didnt click on refresh button its work fine but click on refresh and see the file

I don't understand what you mean. Are you referring to a different bug? Can you elaborate?

PR #47, opened 2023-11-07
Closes Shuffle/Shuffle#1228

0x0elliot:


0x0elliot:
closes https://github.com/Shuffle/Shuffle/issues/1228

PR #51, opened 2024-02-24
added a  method to get the child organizations

satti-hari-krishna-reddy:
`HandleGetSubOrgs` can be used to get all the Child Organizations of the current organization

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -816,7 +816,7 @@ func HandleGetSubOrg(resp http.ResponseWriter , request *http.Request) {
 		orgId = location[4]
 	}
 
-	if strings.Contains(fileId, "?") {
+	if strings.Contains(orgId, "?") {
 		orgId = strings.Split(orgId, "?")[0]
 	}
 
@@ -845,7 +845,7 @@ func HandleGetSubOrg(resp http.ResponseWriter , request *http.Request) {
 		  return
 	}
 
-	subOrgs , err = GetChildOrgs(ctx , orgId)
+	subOrgs , err := GetChildOrgs(ctx , orgId)
 
 	if err != nil {
 		log.Printf("[WARNING] Failed getting sub orgs: %s. Org ID: %s", err, orgId)
--- a/shared.go
+++ b/shared.go
@@ -876,6 +876,7 @@ func HandleGetSubOrg(resp http.ResponseWriter , request *http.Request) {
 			subOrgs = append(subOrgs, OrgMini{
 				Id:         childorg.Id,
 				Name:       childorg.Name,
+				Role:       childorg.Role,
 				CreatorOrg: childorg.CreatorOrg,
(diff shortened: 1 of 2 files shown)


shared.go
```diff
@@ -784,6 +784,89 @@ func HandleGetOrg(resp http.ResponseWriter, request *http.Request) {
...
+	user, err := HandleApiAuthentication(resp, request)
+	
+	if err != nil {
+		log.Printf("[WARNING] Api authentication failed in get org: %s", err)
+		resp.WriteHeader(401)
+		resp.Write([]byte(`{"success": false}`))
+		return
+	}
+
+    orgFound:=false
```
frikky: Indentation please :)

shared.go
```diff
@@ -784,6 +784,89 @@ func HandleGetOrg(resp http.ResponseWriter, request *http.Request) {
...
+		  }
+		}
+	if !orgFound {
+		  log.Printf("[ERROR] User '%s' (%s) isn't a part of org %s (get)", user.Username, user.Id, orgId)
+		  resp.WriteHeader(401)
+	      resp.Write([]byte(`{"success": false, "reason": "User doesn't have access to org"}`))
+		  return
+	}
+
+	subOrgs , err = GetChildOrgs(ctx , orgId)
```
frikky: comma usage is not SPACE COMMA SPACE

text , text2 = bad
text, text2 = good

Space after comma :))

frikky:
Couple of general things:
- Do indentation right. Get a Golang linter.
- Be consistent.

As for the function itself, please answer:
- If we merge this, who can get access to the suborg info?
- Who SHOULD have access to suborg info? Why?
- If inside a suborg (not parent workflow), what should we show?

satti-hari-krishna-reddy:
Access to the suborg info is determined by the authentication process. In the provided code, it checks if the authenticated user is part of the sub organization. If the user is found , they are granted access. Therefore, anyone with valid authentication credentials and is a user of the sub organization can get access to the suborg info.

shared.go
satti-hari-krishna-reddy: i think you are looking at the previous commit not the latest one : )

satti-hari-krishna-reddy:
when there are NO sub orgs found then the code is returning an empty list for now , this is same if we are inside a suborg

frikky:
> Access to the suborg info is determined by the authentication process. In the provided code, it checks if the authenticated user is part of the sub organization. If the user is found , they are granted access. Therefore, anyone with valid authentication credentials and is a user of the sub organization can get access to the suborg info.

To add to this, do:
- Users that are admin in the parent org can see all suborgs of the parent
- Support users can see all suborgs of the parent
- If in a suborg, make it clear what is the parent org. Show parent org only if user is part of parent org or support user

PR #55, opened 2024-03-21
Delete User account feature backend

LalitDeore:


0x0elliot:
okay here is what we would do: i am going to take care of the deletion propagation stuff for you. refer to it to know how someone would implement a feature like that safely.

0x0elliot (changes requested):

db-connector.go
```diff
@@ -4722,6 +4722,50 @@ func SetUser(ctx context.Context, user *User, updateOrg bool) error {
...
+		}
+
+		wrapped := UserWrapper{}
+		err = json.Unmarshal(respBody, &wrapped)
+		if err != nil {
+			return err
+		}
+	} else {
+		key := datastore.NameKey(nameKey, username, nil)
+		err := project.Dbclient.Delete(ctx, key)
```
0x0elliot, line 4756: If deletion is successful, please add a go routine which calls propagateUser like this: propagateUser(*user, true)

0x0elliot (changes requested):
Can you please add just this one line in? I added a few changes that will help you with cross region propagation.

PR #67, opened 2024-05-20
handle no org, create default at login

yashsinghcodes:
This fixes a issue if by any case the user lost any org maybe due to data corruption or lost a default org will be created at the time of there login instead of showcasing this error message `No organization available. Please contact your team, or the shuffle if you think there has been a mistake: support@shuffler.io`


frikky:
Merging as we kind of need this asap, but it's not even close to what we discussed

PR #70, opened 2024-05-21
org fixes for different cases

yashsinghcodes:
Added two new cases:

- If workflow have been built in an org, then users lost access somehow
- If user already points to an org, but that org does not exist, we recreate it.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -10933,45 +10933,72 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
 
 		if userdata.ActiveOrg.Id == "" {
 			if len(userdata.Orgs) == 0 {
-				log.Printf("[WARNING] User %s (%s) has no chosen org. ID: %s, Name: %s. Creating a default org", userdata.Username, userdata.Id, userdata.ActiveOrg.Id, userdata.ActiveOrg.Name)
-				orgSetupName := "default"
-				orgId := uuid.NewV4().String()
-				newOrg := Org{
-					Name:      orgSetupName,
-					Id:        orgId,
-					Org:       orgSetupName,
-					Users:     []User{userdata},
-					Roles:     userdata.Roles,
-					CloudSync: false,
-				}
-
-				err := SetOrg(ctx, newOrg, newOrg.Id)
+				log.Printf("[WARNING] User %s (%s) has no chosen org. ID: %s, Name: %s. Checking for any workflows", userdata.Username, userdata.Id, userdata.ActiveOrg.Id, userdata.ActiveOrg.Name)
+				
+				// case for lost org but there are workflows
+				OrgFound := false
+				workflows, err := GetAllWorkflowsByQuery(ctx, userdata)
 
 				if err != nil {
-					log.Printf("[ERROR] Failed setting default org for the user: %s", userdata.Username)
-				} else {
-					log.Printf("[DEBUG] Successfully created the default org!")
+					log.Printf("[WARNING] Faild to get any workflow")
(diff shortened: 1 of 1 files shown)


shared.go
```diff
@@ -10933,45 +10933,72 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
 
 		if userdata.ActiveOrg.Id == "" {
```
frikky, line 10934: Everything being within this makes it expect that the org is not set in the user. This is most likely wrong

shared.go
```diff
@@ -10933,45 +10933,72 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
...
 			if len(userdata.Orgs) == 0 {
```
frikky, line 10935: Everything in this again does the same

shared.go
```diff
@@ -10933,45 +10933,72 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
...
-				} else {
-					log.Printf("[DEBUG] Successfully created the default org!")
+					log.Printf("[WARNING] Faild to get any workflow")
+				}
 
-					defaultEnv := os.Getenv("ORG_ID")
-					if len(defaultEnv) == 0 {
-						defaultEnv = "Shuffle"
-						log.Printf("[DEBUG] Setting default environment for org to %s", defaultEnv)
+				for _, workflow := range workflows {
```
frikky, line 10946: This is overcomplicating it. The whole point here is that IF a workflow has an org, we just set the user to have that org.  

I also think GetAllWorkflowsByQuery() won't give you much, as that is based on searching in the org, and not the entire instance. THis is obviously not tested

shared.go
```diff
@@ -10933,45 +10933,72 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
...
 					}
+				}
 
-					item := Environment{
-						Name:    defaultEnv,
-						Type:    "onperm",
-						OrgId:   orgId,
-						Default: true,
-						Id:      uuid.NewV4().String(),
+				if !OrgFound {
```
frikky, line 10962: This part is good, but should for sure not be inside all the if statements above

shared.go
```diff
@@ -11013,11 +11040,28 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
 		if !found {
```
frikky, line 11040: Why isn't this the same code that creates a new org on line 10965?

frikky (commented):
Still expects that the database is good. Write code as if random shit is deleted. You can't expect data in user/customer environments to be 100% good, hence you need to write code that fixes every scenario

shared.go
yashsinghcodes: I think it is correct, this "if" is to handle if there is no ActiveOrg for the user.

shared.go
yashsinghcodes: ooh, so should I be looking for Workflow Executions. As there are no other function to get workflow or do you want me to write one from scratch?

shared.go
yashsinghcodes: This is for case if for any case we made a mistake where a user is pointing to an ActiveOrg which user was never a part of so `found` is a boolean which checks if user even is part of the ActiveOrg or not. If not we just look through all the other orgs that users has and in between I wrote a function where we check if the user have a Org Id but that org does not exist we just re-create that old org and pass it on to user.

yashsinghcodes:
Closing this one and re-opening a new fresh one!

PR #72, opened 2024-05-21
backend for the make sso required or optional for org

LalitDeore:


frikky:
@LalitDeore - I don't like the parts where you blatantly copied parts of the original login function that are not SSO related. If they are doing the SSO login, it should FORCE the use of it. 

Merging either way, as this has taken way too much time and I'm gonna fix it on the fly

PR #78, opened 2024-06-14
Billing management and change billing email id 

LalitDeore:


db-connector.go
```diff
@@ -160,6 +162,91 @@ func handleDailyCacheUpdate(executionInfo *ExecutionInfo) *ExecutionInfo {
 	return executionInfo
 }
 
+func sendMailSendgrid(toEmail []string, subject, body string, emailApp bool) error {
```
frikky, line 165: Is it really necessary to re-make the same function in shuffle-shared as well?

It for sure should not be in db-connector as that has to do with data getting/setting

db-connector.go
```diff
@@ -270,6 +357,40 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
...
+
+	org, err := GetOrg(ctx, orgId)
+	if err != nil {
+		log.Printf("[ERROR] Failed getting org in increment: %s", err)
+	}
+
+	log.Printf("Org.Id length is: %v", len(org.Id))
+	if len(org.Id) > 0 {
+		for index, AlertThreshold := range org.Billing.AlertThreshold {
+			if int64(AlertThreshold.Count) < orgStatistics.MonthlyAppExecutions && AlertThreshold.Email_send == false {
```
frikky, line 372: Have you tried running many workflows at the same time with this? This can EASILY run multiple times if you trust GetOrg() to be fast enough, and not a smaller cache object.

db-connector.go
```diff
@@ -270,6 +357,40 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
...
+	//send mail if the app runs more than the set threshold limit
+	ctx := context.Background()
+	orgId := orgStatistics.OrgId
+
+	org, err := GetOrg(ctx, orgId)
+	if err != nil {
+		log.Printf("[ERROR] Failed getting org in increment: %s", err)
+	}
+
+	log.Printf("Org.Id length is: %v", len(org.Id))
```
frikky, line 369: Don't have random prints when we are going to prod

db-connector.go
```diff
@@ -270,6 +357,40 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
 		}
 	}
 
+	//send mail if the app runs more than the set threshold limit
+	ctx := context.Background()
+	orgId := orgStatistics.OrgId
+
+	org, err := GetOrg(ctx, orgId)
+	if err != nil {
+		log.Printf("[ERROR] Failed getting org in increment: %s", err)
```
frikky, line 366: Shouldn't it return here? I think all this checking should have its own function

shared.go
```diff
@@ -11332,6 +11362,96 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
 
 }
 
+func changeBillingEmail(ctx context.Context, org *Org, NewEmail string) (*Org, error) {
```
frikky, line 11365: Shouldn't this function be on cloud only? Why is it in shuffle-shared?

No stripe necessary onprem :)

frikky (commented):
Look into this please

db-connector.go
LalitDeore: "sendMailSendgrid" function present in cloud_specific.go file is not accessible in this package But yes i have to move this function in shared.go

LalitDeore: "sendMailSendgrid" function present in the cloud_specific.go was not accessible in this package but yes i have to move this to the shared.go file.

shared.go
LalitDeore: I am thinking of running it only when Project.Environment == 'cloud'. Will that work?

db-connector.go
LalitDeore: No, I haven't try this with running multiple workflow at the same time will it become slow because of GetOrg run multiple times?

frikky: What matters is making sure another email wasn't sent for the same thing :)

shared.go
frikky: If it's only necessary on cloud, we only have it on cloud. Move it to the go-app/backend/cloud_specific.go please :)

LalitDeore: If i move it to the cloud_specific.go than this "changeBillingEmail" function will no be accessible in shuffle-shared package.

frikky: > If i move it to the cloud_specific.go than this "changeBillingEmail" function will no be accessible in shuffle-shared package.

Why is that wrong?

db-connector.go
LalitDeore: It will not send multiple emails for the same threshold. Once an email is sent for a particular threshold, the email_send value is set to true for that threshold. After that, if the email_send value is true, it will not send the email again.

shared.go
LalitDeore: I tried to place it in cloud_specific.go, but this file is outside of the shuffle -shared package. Because of this, I am unable to access this function in shared.go.

db-connector.go
frikky: > It will not send multiple emails for the same threshold. Once an email is sent for a particular threshold, the email_send value is set to true for that threshold. After that, if the email_send value is true, it will not send the email again.

Yes, but I'm talking about the speed of this. You are not handling relatively normal cases where many workflows run at once and are trying to update the stats at the same time. 

Just add a simple local cache that it checks (like you did with cache for MFA IDs), as Org is a larger entity, and will load from cache slower. You can NOT trust the DB to be fast enough to handle this at all.

shared.go
frikky: > I tried to place it in cloud_specific.go, but this file is outside of the shuffle -shared package. Because of this, I am unable to access this function in shared.go.

But why do you need the function in shared.go?

LalitDeore: The function to change the billing email is added in HandleEditOrg. It updates the billing email in Stripe and also in the organization. Should i create new endpoint for this, so it will do all stuff in cloud_specific.go only?

frikky: > The function to change the billing email is added in HandleEditOrg. It updates the billing email in Stripe and also in the organization. Should i create new endpoint for this, so it will do all stuff in cloud_specific.go only?

Ah I see! Good question 🤔 

I think we indeed should do this with a new endpoint on cloud_specific.go

LalitDeore: > > The function to change the billing email is added in HandleEditOrg. It updates the billing email in Stripe and also in the organization. Should i create new endpoint for this, so it will do all stuff in cloud_specific.go only?
> 
> Ah I see! Good question 🤔
> 
> I think we indeed should do this with a new endpoint on cloud_specific.go. Maybe have like an -XPUT /api/v1/orgs/{orgid}/billing?

yes will make this new endpoint.

PR #88, opened 2024-06-25
manage user's consultation hour

LalitDeore:


LalitDeore:
In this PR, I have also added a function to send private training requests email.

shared.go
```diff
@@ -28048,3 +28051,100 @@ func GetWorkflowValidation(resp http.ResponseWriter, request *http.Request) {
...
+	ctx := GetContext(request)
+	org, err := GetOrg(ctx, tmpData.OrgId)
+	if err != nil {
+		log.Printf("[ERROR] Failed getting org %s: %s", tmpData.OrgId, err)
+		resp.WriteHeader(http.StatusBadRequest)
+		resp.Write([]byte(`{"success": false}`))
+		return
+	}
+
+	email := []string{org.Org}
```
0x0elliot, line 28122: Isn't this assuming that org.Org contains a valid email? Many org name's are NOT emails since they are changeable. I would have instead tried to take the admins and emailed them one by one.

LalitDeore: Will the userName always be an email? Can I use that field?

0x0elliot: yes, user.Username is always an email :)

PR #90, opened 2024-06-28
fix the openid login issue

LalitDeore:


0x0elliot:
@LalitDeore why are you org name as the preferred email lol

PR #95, opened 2024-07-30
fix for self signed certificate

yashsinghcodes:
Manually appending the cert to the trusted CA as mounting the drive there would overwrite stuffs which could be bad. So more control over appending the certs better. Will also make the changes in docker-compose in Shuffle.


frikky:
Keeping this open until you got what we talked about today fixed.

PR #98, opened 2024-08-07
[fix]: subflow execution with .#1

yashsinghcodes:
fix for Shuffle/Shuffle#1463

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -159,6 +159,12 @@ func randStr(strSize int, randType string) string {
 	return string(bytes)
 }
 
+func isLoop(arg string) bool {
+    pattern := `(^|\.)(#(\d+-\d+)?($|\.))`
+    re := regexp.MustCompile(pattern)
+    return strings.Contains(arg, "$") && re.MatchString(arg)
+}
+
 func HandleSet2fa(resp http.ResponseWriter, request *http.Request) {
 	cors := HandleCors(resp, request)
 	if cors {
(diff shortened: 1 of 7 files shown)


yashsinghcodes:
also fixes the Shuffle/Shuffle#1492 for shuffle-shared as it was required to actually test my file.

yashsinghcodes:
@frikky please have a look :)

shared_test.go
```diff
@@ -0,0 +1,26 @@
...
+    handlers := []struct {
+        arg         string
+        expected    bool
+    }{
+        {"$exec.#1-2", true},
+        {"$exec.#.value.#1", true},
+        {"$exec.#1", false},
+        {"$exec", false},
+        {"$exec.#1.value.#2", false},
+        {"$start_node.#", true},
```
frikky, line 17: Remember some newline tests ;)

shared.go
```diff
@@ -159,6 +159,15 @@ func randStr(strSize int, randType string) string {
 	return string(bytes)
 }
 
+func isLoop(arg string) bool {
+    if strings.Contains(arg,"$") && strings.Contains(arg, ".#") {
+        pattern := `(^|\.)(#(\d+-\d+)?($|\.))`
```
frikky, line 164: Can we add another if inside the other if that checks for string contains  ".#." or string endswith(".#") and just return true if it is?

I want to use the regex as little as possible :)

PR #99, opened 2024-08-09
[fix]: wrong condition for proxy handling

yashsinghcodes:
Due to the wrong condition handling the client which we use does not uses the SHUFFLE_INTERNAL_PROXY, the HTTP_PROXY works because linux uses that as a default one.

fix for Shuffle/Shuffle#1474


frikky:
This isn't fixed after our last talk right? It will do proxy overrides the wrong way I believe.

PR #101, opened 2024-08-16
fix for health api

yashsinghcodes:
As the workflow from which we get our actions does not seem to exist anymore, I just used the other workflow which was commented out as they seem to have similar set of actions. Quick fix for the Health Api and will be adding more deeper test in future.


yashsinghcodes:
without change:
Screenshot 2024-08-17 013459

with changes:
Screenshot 2024-08-17 013824

PR #102, opened 2024-08-19
sent alert email to each admin in org.

LalitDeore:
Added logic to sent alert email to each admin present in the org when app run is more than the set alert threshold.

### DIFF
--- a/db-connector.go
+++ b/db-connector.go
@@ -296,27 +296,35 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
 		return orgStatistics
 	}
 
+	//send mail if the app runs more than the set threshold limit
+	emailSend := false
 	if len(org.Id) > 0 {
 		for index, AlertThreshold := range org.Billing.AlertThreshold {
 			if int64(AlertThreshold.Count) < orgStatistics.MonthlyAppExecutions && AlertThreshold.Email_send == false {
-				mailbody := Mailcheck{
-					Targets: []string{org.Org},
-					Subject: "You have reached the threshold limit of app executions.",
-					Body:    fmt.Sprintf("You have reached the threshold limit of %v percent Or %v app executions run. Please login to shuffle and check it.", AlertThreshold.Percentage, AlertThreshold.Count),
+				for _, user := range org.Users {
+					if user.Role == "admin" {
+						mailbody := Mailcheck{
+							Targets: []string{user.Username},
+							Subject: "You have reached the threshold limit of app executions",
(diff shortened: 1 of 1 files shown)


db-connector.go
```diff
@@ -296,27 +296,35 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
...
-				mailbody := Mailcheck{
-					Targets: []string{org.Org},
-					Subject: "You have reached the threshold limit of app executions.",
-					Body:    fmt.Sprintf("You have reached the threshold limit of %v percent Or %v app executions run. Please login to shuffle and check it.", AlertThreshold.Percentage, AlertThreshold.Count),
-				}
-				err = sendMailSendgrid(mailbody.Targets, mailbody.Subject, mailbody.Body, false)
-				if err != nil {
-					log.Printf("[ERROR] Failed sending alert mail in increment: %s", err)
+				for _, user := range org.Users {
+					if user.Role == "admin" {
```
frikky, line 305: Just for code readability, this construct is better due to indentation levels (opposite):
```
if user.Role != "admin" {
   continue
}

// actual code here :)

```

db-connector.go
```diff
@@ -296,27 +296,35 @@ func HandleIncrement(dataType string, orgStatistics *ExecutionInfo, increment ui
...
+						err = sendMailSendgrid(mailbody.Targets, mailbody.Subject, mailbody.Body, false)
+						if err != nil {
+							log.Printf("[ERROR] Failed sending alert mail in increment: %s", err)
+						} else {
+							emailSend = true
+						}
+					}
 				}
-				if err == nil {
+				if emailSend {
```
frikky, line 319: I'm kind of worried if this is fast enough to not send duplicates. May need to set cache.

db-connector.go
LalitDeore: I will make this change

db-connector.go
LalitDeore: "emailSend" is just a variable that checks whether any email has been sent or not. Could you please tell me why we need to set this single variable in the cache? We're not retrieving it from the database; we have just assigned it locally. Initially, it is set to false, and if any email is sent, we simply update it to true even if one mail is sent.

frikky: I didn't mean this part directly - just the decisionmaker for whether to send another email or not in general. 

The point: If 100 workflows at the same time, how can you make sure that there is literally 0% chance even if they happen around the same time that emails are sent more than once? In short: Is the time between apps running and the organization set in cache fast enough? I'd say most likely not, as the data for an org is larger and would need to be transferred over some network connection.

I'll merge this for now, but I want you to be aware that this WILL occur, and someone will at some point have a problem with it.

PR #103, opened 2024-08-21
changed, to use isLoop only where was needed

yashsinghcodes:
current status:
Screenshot 2024-08-21 150438

fix:

https://github.com/user-attachments/assets/94fb5d22-3dd5-4e31-acb1-b2d49db75a7b


yashsinghcodes:
Current know egdecases for this implementation:
- New functions like for Tenzir
- Backend functions talking to themself

PR #105, opened 2024-09-02
Fix SSO redirect url assign issue

LalitDeore:
I found a bug related to assigning the redirect URL. It was incorrectly assigning the redirect URL as SSO_REDIRECT_URL, which is the environment variable for our frontend URL where the user will be redirected after login. In this function, we should assign the redirect URL as our backend endpoint for the redirection not the fronend. Also, added some checks  for handling all cases.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -12773,7 +12773,7 @@ func GetOpenIdUrl(request *http.Request, org Org) string {
 		//check if base url exist if exist then assign the base url. This is for local testing when request.Host is is not "shuffle-backend"
 		if project.Environment != "cloud" && len(os.Getenv("BASE_URL")) > 0 {
 			redirectUrl = url.QueryEscape(fmt.Sprintf("%s/api/v1/login_openid", os.Getenv("BASE_URL")))
-		} else {
+		} else if project.Environment != "cloud" {
 			//if base url not exist then assign hardcoded url for the onprem, user should not reach here but in case not set the base url hardcode it.
 			redirectUrl = url.QueryEscape(fmt.Sprintf("http://localhost:5001/api/v1/login_openid"))
 		}
--- a/shared.go
+++ b/shared.go
@@ -18922,7 +18922,8 @@ func HandleSSO(resp http.ResponseWriter, request *http.Request) {
 	}
 
 	if len(backendUrl) > 0 {
-		redirectUrl = fmt.Sprintf("%s/workflows", backendUrl)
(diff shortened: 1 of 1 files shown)


shared.go
```diff
@@ -12766,13 +12766,23 @@ func GetOpenIdUrl(request *http.Request, org Org) string {
...
+	//Redirect url for onprem
 	if project.Environment != "cloud" && strings.Contains(request.Host, "shuffle-backend") && !strings.Contains(os.Getenv("BASE_URL"), "shuffle-backend") {
 		redirectUrl = url.QueryEscape(fmt.Sprintf("%s/api/v1/login_openid", os.Getenv("BASE_URL")))
+	} else {
+		//check if base url exist if exist then assign the base url. This is for local testing when request.Host is is not "shuffle-backend"
+		if project.Environment != "cloud" && len(os.Getenv("BASE_URL")) > 0 {
+			redirectUrl = url.QueryEscape(fmt.Sprintf("%s/api/v1/login_openid", os.Getenv("BASE_URL")))
+		} else {
+			//if base url not exist then assign hardcoded url for the onprem, user should not reach here but in case not set the base url hardcode it.
+			redirectUrl = url.QueryEscape(fmt.Sprintf("http://localhost:5001/api/v1/login_openid"))
```
frikky, line 12778: Uhm, doesn't cloud go straight here? So cloud SSO would use localhost somehow..?

shared.go
```diff
@@ -12773,7 +12773,7 @@ func GetOpenIdUrl(request *http.Request, org Org) string {
 		//check if base url exist if exist then assign the base url. This is for local testing when request.Host is is not "shuffle-backend"
 		if project.Environment != "cloud" && len(os.Getenv("BASE_URL")) > 0 {
 			redirectUrl = url.QueryEscape(fmt.Sprintf("%s/api/v1/login_openid", os.Getenv("BASE_URL")))
```
frikky, line 12775: Why is this based on `BASE_URL` and not `SSO_REDIRECT_URL`? Is that handled somewhere else? Is SSO and OpenID connect handled the same way?

LalitDeore:
In this commit ( https://github.com/Shuffle/shuffle-shared/pull/105/commits/5836d4448ca02f27143ef4a828b643ccd9127fbd ) added one more change. As SSO_REDIRECT_URL already contain /workflow endpoint by default like httt://localhost:3001/workflow so we don't need to add /workflow endpoint in it else url become like http://localhost:3001/workflow/workflow and dosn't redirect properly.

PR #107, opened 2024-09-24
Fix the SSO redirection issue when switching organizations and resolve the issue with managing user roles.

LalitDeore:
Frontend pr: https://github.com/Shuffle/shaffuru/pull/374

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -1449,6 +1449,9 @@ func HandleLogout(resp http.ResponseWriter, request *http.Request) {
 	DeleteCache(ctx, fmt.Sprintf("session_%s", userInfo.Session))
 	DeleteCache(ctx, userInfo.Session)
 
+	//store user's last session so we can force sso when user's session change.
+	userInfo.UsersLastSession = userInfo.Session
+
 	userInfo.Session = ""
 	userInfo.ValidatedSessionOrgs = []string{}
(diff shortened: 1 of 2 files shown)


shared.go
```diff
@@ -10828,8 +10834,7 @@ func HandleChangeUserOrg(resp http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	//if org.SSOConfig.SSORequired == true && !ArrayContains(user.ValidatedSessionOrgs, tmpData.OrgId) && user.SupportAccess == false {
-	if org.SSOConfig.SSORequired == true && !ArrayContains(user.ValidatedSessionOrgs, tmpData.OrgId) && user.SupportAccess == false {
+	if (org.SSOConfig.SSORequired == true && user.UsersLastSession != user.Session && user.SupportAccess == false) || tmpData.SSOTest {
```
0x0elliot, line 10837: Let's test this on call today. I want to verify how you're trying to handle auth here. As in:

1. Do you wish that the user must always sign in with SSO first and then be able to do any org related operations?
2. Or do you wish for SSO to be just another layer? (This doesn't make sense to me).
3. Or, do you wish to generate a shuffle account for the user automatically when they login through SSO?

I am still catching up with what you have been doing here so don't mind me being a little naive. These questions are important to merge in everything securely.

PR #111, opened 2024-10-03
FIX FOR SUBFLOW HALTING ISSUE

yashsinghcodes:
A short fix for subflow halting issue where the subflow status was halted on waiting even after getting back the results.

The Solution is we get all the `parentSubflowResult` which is the execution of the subflow -> after we get it we just compare if we have gotten back all the results from our invocation.  We track the finished subflow with `finishedSubflows`

@frikky let me know if this sounds good!


yashsinghcodes:
Sorry for the formatting, I did used `gofmt`

shared.go
```diff
@@ -14064,6 +14063,14 @@ func updateExecutionParent(ctx context.Context, executionParent, returnValue, pa
 				newResults = append(newResults, res)
 			}
 
+			log.Printf("[INFO][%s] TOTAL FINISHED SUBFLOWS: %d/%d", subflowExecutionId, len(parentSubflowResult), len(newResults))
```
frikky, line 14066: Isn't this counting how many for the parent are done? That means it shouldn't be using subflowExecutionId for the print. Removes searchability

shared.go
```diff
@@ -14064,6 +14063,14 @@ func updateExecutionParent(ctx context.Context, executionParent, returnValue, pa
...
+
+			// Can it be if this and also status = "WAITING"?
+			if len(parentSubflowResult) == finishedSubflows {
```
frikky, line 14069: I suggest doing >= for these kinds of result handlers, as there may be times duplicate counts happen, and you go above. It won't be as accurate, but it's better than possibilities of failure

shared.go
```diff
@@ -14064,6 +14063,14 @@ func updateExecutionParent(ctx context.Context, executionParent, returnValue, pa
...
+				log.Printf("[INFO][%s] ALL THE SUBFLOW GOT THE RESULT BACK SO UPATING THE STATUS TO SUCCESS")
+				foundResult.Status = "SUCCESS"
```
frikky, line 14071: I checked the existing shared.go which was this the below code. Are we just re-implementing that? Did you check the commit that commented the section? I'm curious how that came about.

```
					foundResult.Status = "SUCCESS"

					if result.CompletedAt == 0 {
						result.CompletedAt = time.Now().Unix()*1000
					}
```

frikky (commented):
Check my code comments :)

frikky:
One more thing: since we are updating a random object `foundResult.success`, can we be sure it will run the `ranUpdate` function? 

https://github.com/Shuffle/shuffle-shared/blob/5a30d86b73f3bf44d20a7446161364872288b53c/shared.go#L14081

PR #117, opened 2024-10-15
added some more category&feilds

ayush0033:
Added blank fields, updated some descriptions to suggest more information, and included additional keywords.


frikky:
> Added blank fields, updated some descriptions to suggest more information, and included additional keywords.

I think you also reintroduced usecases we don't want with this PR. Please change it to have fewer usecases (like current one), but with a lot more details for each one.

frikky:
Closing as I got no feedback on this - please reopen with fixes

PR #119, opened 2024-10-23
checking if the workflow exist before saving after validation

yashsinghcodes:
Quick fix after yesterday outage where the workflow was keep recreating due to the validation part.


shared.go
```diff
@@ -29289,6 +29289,14 @@ func checkExecutionStatus(ctx context.Context, exec *WorkflowExecution) *Workflo
 	workflowChanged = true
 	workflow.Validation.ValidationRan = true
 	workflow.Validation.ExecutionId = exec.ExecutionId
+
+	// Check if the workflow exist or not before saving (fix after 22/10/24 outage)
+        _, err = GetWorkflow(ctx, workflow.ID)
```
frikky, line 29294: Why are we getting the workflow, when we should be removing the SetWorkflow part entirely? 

This makes very very little sense. This would require perfect deletion of Cache + Database timing down to the millisecond (or even cllser) with how the outage was

frikky (commented):
Just remove what should be removed. Don't add even more validation

frikky:
@yashsinghcodes don't do `git commit -am`, and don't `git add *` please. 

You pushed things related to app_upload 👎

yashsinghcodes:
@frikky This is the dumb mistake from my side I thought I have no PR open from [fix-outage-issue](https://github.com/yashsinghcodes/shuffle-shared/tree/fix-outage-issue) branch. I will resolve this ASAP. Thank you so much for pointing it out.

frikky:
Fixed it already

PR #120, opened 2024-10-28
Fix - SSO redirection issue

LalitDeore:
Default frontend redirect url for SSO is http://localhost:3001/workflows if SSO_REDIRECT_URL exist than use SSO_REDIRECT_URL/workflows.


shared.go
```diff
@@ -19370,8 +19373,7 @@ func HandleSSO(resp http.ResponseWriter, request *http.Request) {
 	}
 
 	if len(backendUrl) > 0 {
-		//we don't need to add /workflow path in backend url as backend url is SSO_REDIRECT_URL and it is already pointing to /workflow by default.
-		redirectUrl = backendUrl
+		redirectUrl = fmt.Sprintf("%s/workflows", backendUrl)
```
frikky, line 19376: This doesn't take into account removing the /api part of the path in any way. We NEED to check if the URL should be modified or not.

This is just randomly guessing that the /workflows is NOT there for some reason. Add validation to ENSURE the URL is ALWAYS correct.

frikky:
@LalitDeore are we merging this? 🤔

frikky:
Closing as no reply

PR #121, opened 2024-11-04
Fix so checkAction will be able to validate if the next action should be skipped or not

yashsinghcodes:
The approach is pretty simple we skip the actions if all the parents are skipped.

Now https://github.com/Shuffle/shuffle-shared/commit/3621997088401585ca4614b7bfedf32de67d2785 fixes the issue where it was not related to child node skip but where the decide execution was not able to get the next action. Adding a validation with `CheckNextAction` at the main `DecideExecution` should already fix the problem as I found out during my tests but by making `CheckNextAction` itself skips the action saves a lot of time in my opinion.

@frikky let me know what you think if there are any mistake I'll try to resolve and retest ASAP. Thanks for all the help on this really learnt a lot :pray:

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -23626,27 +23626,32 @@ func CheckNextActions(ctx context.Context, workflowExecution *WorkflowExecution)
 
 	for index, actionId := range nextActions {
 		skippedParents := 0
+
+		if _, ok := parents[actionId]; !ok {
+			continue
+		}
+
 		for _, parent := range parents[actionId] {
 			_, result := GetActionResult(ctx, *workflowExecution, parent)
 			if result.Status == "SKIPPED" {
 				skippedParents += 1
 			}
 		}
-		if skippedParents >= len(parents[actionId]) {
-			if actionId != workflowExecution.Workflow.Start {
-				for _, action := range workflowExecution.Workflow.Actions {
-					if actionId != action.ID {
-						continue
-					}
-					foundAction := GetAction(*workflowExecution, actionId, action.Environment)
-					err := ActionSkip(foundAction, *workflowExecution, parents[actionId])
-					if err != nil {
-						log.Printf("ERROR %s", err)
-					}
-					copy(nextActions[index:], nextActions[index+1:])
-					nextActions[len(nextActions)-1] = ""
-					nextActions = nextActions[:len(nextActions)-1]
+
+		if skippedParents >= len(parents[actionId]) && actionId != workflowExecution.Start {
+			for _, action := range workflowExecution.Workflow.Actions {
+				if actionId != action.ID {
+					continue
 				}
+
+				foundAction := GetAction(*workflowExecution, actionId, action.Environment)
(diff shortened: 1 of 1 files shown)


yashsinghcodes:
Please if in future while merging also merge the https://github.com/Shuffle/shaffuru/pull/390 as it breaks one of the condition which could create a problem.

shared.go
```diff
@@ -23629,11 +23624,104 @@ func CheckNextActions(ctx context.Context, workflowExecution *WorkflowExecution)
 		}
 	*/
 
-
+	for index, actionId := range nextActions {
+		skippedParents := 0
+		for _, parent := range parents[actionId] {
+			_, result := GetActionResult(ctx, *workflowExecution, parent)
+			if result.Status == "SKIPPED" {
```
frikky, line 23631: Have you verified this works, both with a few skipped and a lot of them, along with some success, some failure etc? 

This feels wrong, but I'm not sure why.

shared.go
```diff
@@ -23629,11 +23624,104 @@ func CheckNextActions(ctx context.Context, workflowExecution *WorkflowExecution)
...
+					if actionId != action.ID {
+						continue
+					}
+					foundAction := GetAction(*workflowExecution, actionId, action.Environment)
+					err := ActionSkip(foundAction, *workflowExecution, parents[actionId])
+					if err != nil {
+						log.Printf("ERROR %s", err)
+					}
+					copy(nextActions[index:], nextActions[index+1:])
+					nextActions[len(nextActions)-1] = ""
```
frikky, line 23647: This section is overcomplicated, and I'm worried about indexing issues. Did you test it well, both with a lot of items and 0? This seems like it can crash if there is only 1 item.

I typically make a new array entirely for stuff like this, as index modification is awful. Such as: Why would it be necessary to set a value to "" if you are just about to delete it? That doesn't seem necessary.

shared.go
```diff
@@ -23629,11 +23624,104 @@ func CheckNextActions(ctx context.Context, workflowExecution *WorkflowExecution)
...
+	newResult := ActionResult{
+		Action:        foundAction,
+		ExecutionId:   exec.ExecutionId,
+		Authorization: exec.Authorization,
+		Result:        fmt.Sprintf(`{"success": false, "reason": "Skipped because of previous node - %d" - %v}`, len(parent), parent),
+		StartedAt:     0,
+		CompletedAt:   0,
+		Status:        "SKIPPED",
+	}
+	resultData, err := json.Marshal(newResult)
```
frikky, line 23667: Add some space between each section. This is hard to read. 

The simplest "rule" for easy readability and space is to have at least one newline at the end of a section, such as:

`if x == y {

}

if x2 == y2 {

}
```

frikky (commented):
fix pls

frikky:
@yashsinghcodes tell me when it's done please :)

yashsinghcodes:
@frikky resolved the conflicts. Ready for merge.

PR #122, opened 2024-11-05
[fix]: not allowing branches to point to themself

yashsinghcodes:


db-connector.go
```diff
@@ -8154,6 +8154,9 @@ func FixWorkflowPosition(ctx context.Context, workflow Workflow) Workflow {
 		if branch.ID == "" {
 			workflow.Branches[index].ID = uuid.NewV4().String()
 		}
+               if branch.DestinationID == branch.SourceID {
+                        workflow.Branches = append(workflow.Branches[:index], workflow.Branches[index+1:]...)
```
frikky, line 8158: Same issue here: can't this crash due to indexing issues?

Please verify in the other script and come back here to verify.

yashsinghcodes: I have verified, it works for all the cases.

frikky:
I guess we use it then :)

PR #143, opened 2025-01-29
feat: Improvement over health API

yashsinghcodes:


yashsinghcodes:
Please make sure to define a new env variable called SHUFFLE_BACKEND_VERSION to get the records in health api logs.

health.go
```diff
@@ -893,18 +885,57 @@ func fixOpensearch() error {
 	return nil
 }
 
+func FixSubflow(ctx context.Context, workflowId string) error {
```
frikky, line 888: I'm sorry, what is happening here? Why do we have a custom fixSubflow function in the API, which SETS the workflow in the database differently?

If you were just returning the workflow to be used in the parent it's ok, but SETTING, then GETTING it again doesn't really do that. 

Please explain. This makes no sense to me

health.go
```diff
@@ -1209,7 +1237,8 @@ func InitOpsWorkflow(apiKey string, OrgId string) (string, error) {
 	if project.Environment == "cloud" {
 		// url := "https://shuffler.io/api/v1/workflows/602c7cf5-500e-4bd1-8a97-aa5bc8a554e6"
 		// url := "https://shuffler.io/api/v1/workflows/7b729319-b395-4ba3-b497-d8246da67b1c"
-		url := "https://shuffler.io/api/v1/workflows/412256ca-ce62-4d20-9e55-1491548349e1"
+		//url := "https://shuffler.io/api/v1/workflows/412256ca-ce62-4d20-9e55-1491548349e1"
+		url := "http://shuffler.io/api/v1/workflows/ae89a788-a26b-4866-8a0b-ce0b31d354ea"
```
frikky, line 1241: Why do we keep changing URLs? Do you make a new one every time?

health.go
yashsinghcodes: I had to set it this way because we use the API call to the `<workflowID>/execute` endpoint to run the workflow. So, if I modify any field in the workflow, I need to save the workflow. However, you're right—there is a redundant call to GetWorkflow, which I will fix.

I also just realized that instead of using the SetWorkflow function, I should use the API. While we know that SetWorkflow works, we are not actually verifying whether sending JSON to the save API functions correctly (which should test both the SetWorkflow as well as json handling). I'll update the code accordingly.

health.go
yashsinghcodes: The main reason I changed it was that I didn’t want to interfere with the actual public version while our health API was still using it. Editing or adding a new action without proper code could cause false failures. So, I created my own version, but soon I realized that I had made a lot of changes and decided to continue with it.

No, we don’t create a new one every time. This time, we added subflows, which is why I created a new one. I'll remove all the others and move forward with this.

frikky: oki. 

Btw; https URL pls :)

frikky:
Should we merge this? Is it ready?

frikky:
> Please make sure to define a new env variable called SHUFFLE_BACKEND_VERSION to get the records in health api logs.

Send a PR with this updated in every single one. This NEEDS to be dynamic.

yashsinghcodes:
Can we track the version with the deployment time. It will be pretty easy to implement onto the deployment script, so don't have to update the version manually and we can track it with the deployment date.

@frikky

yashsinghcodes:
New Things Added:
- Track execution time
- Make use of the API rather than directly calling the function.

frikky:
> Can we track the version with the deployment time. It will be pretty easy to implement onto the deployment script, so don't have to update the version manually and we can track it with the deployment date.
> 
> @frikky

That is literally what I said. Dynamic. Just use the revision thing Aditya already made.

health.go
```diff
@@ -485,13 +485,6 @@ func RunOpsHealthCheck(resp http.ResponseWriter, request *http.Request) {
 		apiKey = org.Users[validIndex].ApiKey
 	}
 
-	if len(apiKey) == 0 || len(orgId) == 0 {
```
frikky, line 488: What happened to this code? Lol

health.go
```diff
@@ -893,18 +885,104 @@ func fixOpensearch() error {
 	return nil
 }
 
+func fixSubflow(ctx context.Context, workflow *Workflow) error {
```
frikky, line 888: Make the name of the function more specific please. This tells me nothing 😆

health.go
```diff
@@ -893,18 +885,104 @@ func fixOpensearch() error {
...
+					parameter.Value = workflow.ID
+				}
+				if parameter.Name == "startnode" {
+					parameter.Value = subflowActionId
+				}
+			}
+		}
+	}
+
+	baseUrl := os.Getenv("SHUFFLE_CLOUDRUN_URL")
```
frikky, line 913: Write these opposite. Again; SANE DEFAULTS.

Set default to shuffler.io -> override with cloudrun URL

Further; does this code run onprem? Is this handled then, as to point to the right system?

structs.go
```diff
@@ -3965,7 +3965,10 @@ type AppHealth struct {
 type WorkflowHealth struct {
 	Create             bool   `json:"create"`
 	Run                bool   `json:"run"`
+	BackendVersion	   string `json:"backend_version"`
```
frikky, line 3968: Waiting for changes to the deploy files (:

frikky (commented):
More reviews yay

health.go
yashsinghcodes: Done!

Handled the on-prem case as well, tested and working.

health.go
yashsinghcodes: Done! Changed something which make it more specific.

health.go
yashsinghcodes: So, this was the code which was restricting anyone to access the stats. It use to ask for `SHUFFLE_OPS_DASHBOARD_APIKEY or SHUFFLE_OPS_DASHBOARD_ORG not set. Please set these to use this feature!` which doesn't really make sense.

structs.go
yashsinghcodes: Deploy files as in? Would love to get it done, just need more context.

health.go
```diff
@@ -910,12 +907,20 @@ func fixSubflow(ctx context.Context, workflow *Workflow) error {
 		}
 	}
 
-	baseUrl := os.Getenv("SHUFFLE_CLOUDRUN_URL")
-	if len(baseUrl) == 0 {
+	baseUrl := "https://shuffler.io"
+	if len(os.Getenv("SHUFFLE_CLOUDRUN_URL")) > 0 {
```
frikky, line 911: Always use SHUFFLE_CLOUDRUN_URL last, as that makes it possible to use it as an onprem environment variable to override URLs in general.

health.go
```diff
@@ -910,12 +907,20 @@ func fixSubflow(ctx context.Context, workflow *Workflow) error {
...
 		log.Printf("[DEBUG] Base url not set. Setting to default")
-		baseUrl = "https://shuffler.io"
+		baseUrl = os.Getenv("SHUFFLE_CLOUDRUN_URL")
 	}
 
+	if project.Environment == "onprem" {
+		log.Printf("[DEBUG] Onprem environment. Setting base url to localhost")
+		// This will work as the health will be handled in the backend itself.
+		// so localhost just works. (Tested)
+		baseUrl = "http://localhost:5001"
```
frikky, line 920: Is there a way to change the port that the backend runs on? Is BASE_URL relevant? Do we need to load balance onprem requests?

Please think about why your code exists, and how it fits in the architecture.

health.go
```diff
@@ -893,18 +885,113 @@ func fixOpensearch() error {
 	return nil
 }
 
+func fixSubflowParameters(ctx context.Context, workflow *Workflow, apiKey string, orgId string) error {
```
frikky, line 888: Is this a generic function we can pass in a workflow into and it will fix parameters at all times, or is it specifically for health?

If it's for health, please make that clear in the name, as while grepping for relevant functions, I may end up just using it (:

health.go
```diff
@@ -893,18 +885,113 @@ func fixOpensearch() error {
...
+			baseUrl = os.Getenv("BASE_URL")
+		}
+	}
+
+	if len(os.Getenv("SHUFFLE_CLOUDRUN_URL")) > 0 {
+		log.Printf("[DEBUG] Base url not set. Setting to default")
+		baseUrl = os.Getenv("SHUFFLE_CLOUDRUN_URL")
+	}
+
+	req, err := http.NewRequest("PUT", baseUrl+"/api/v1/workflows/"+workflow.ID+"?skip_save=true", nil)
```
frikky, line 928: Make this more readable. Just make a variable for the URL and newline each item. Hard to come in here and get the point without that.

Why is it a PUT request without any data? 🤔

health.go
```diff
@@ -893,18 +885,113 @@ func fixOpensearch() error {
...
 func RunOpsWorkflow(apiKey string, orgId string, cloudRunUrl string) (WorkflowHealth, error) {
 	// run workflow with id 602c7cf5-500e-4bd1-8a97-aa5bc8a554e6
 	ctx := context.Background()
 
 	workflowHealth := WorkflowHealth{
 		Create:      false,
+		BackendVersion: os.Getenv("SHUFFLE_BACKEND_VERSION"),
 		Run:         false,
 		RunFinished: false,
+		ExecutionTook: 0,
```
frikky, line 989: This is a nice touch!

frikky (commented):
Minor things. Merge time?

health.go
yashsinghcodes: It is specifically for the health API. Let me fix the name of the functions to make them more clear :)

health.go
yashsinghcodes: We are actually defining the body in the later part of the code, this is just being use to create new `Request` type.
https://github.com/Shuffle/shuffle-shared/pull/143/files/a9d13e1dae8f0fa3863f10e982a4191dacf46c2f#diff-46ab8d7c52533584f96ba53eaeadd8cf52839f1c5d175c059ceb909c2c3e0715R947

health.go
```diff
@@ -949,18 +941,46 @@ func fixOpensearch() error {
 	return nil
 }
 
+func fixHealthSubflowParameters(ctx context.Context, workflow *Workflow) (Workflow, error) {
+	subflowActionId := ""
+	for _, action := range workflow.Actions {
+		if action.Label == "call_subflow" {
+			subflowActionId = action.ID
```
frikky, line 948: When you look for a single item in a list, ALWAYS break when you find it.

health.go
```diff
@@ -949,18 +941,46 @@ func fixOpensearch() error {
...
+		}
+	}
+
+
+	for i := range workflow.Triggers {
+		if workflow.Triggers[i].AppName == "Shuffle Workflow" {
```
frikky, line 954: Same here; when looking for a single item do both:
1. Opposite search and continue (less indents)
2. Break when you find the item

frikky (commented):
Looking good! Just minor optimizations left.

PR #145, opened 2025-02-03
[feature] - search multitenant workflow run search in parent org

LalitDeore:


shared.go
```diff
@@ -28483,6 +28483,104 @@ func HandleWorkflowRunSearch(resp http.ResponseWriter, request *http.Request) {
...
+		suborgs, err := GetAllChildOrgs(ctx, user.ActiveOrg.Id)
+		if err != nil {
+			log.Printf("[WARNING] Failed getting suborgs for org %s: %s", user.ActiveOrg.Id, err)
+			resp.WriteHeader(400)
+			resp.Write([]byte(`{"success": false}`))
+			return
+		}
+
+		// Limit to max 50 suborgs
+		if len(suborgs) > 50 {
```
frikky, line 28497: Minor unimportant optimization: Shouldn't the check for the user existing in an org happen earlier? Inside the goroutine is "efficient", but it can randomly turn into search e.g. 27 orgs instead of 50, as the cleanup happens before we check if they should be able to see it.

shared.go
```diff
@@ -28483,6 +28483,104 @@ func HandleWorkflowRunSearch(resp http.ResponseWriter, request *http.Request) {
 		Cursor:  cursor,
 	}
 
+	//Get workflow run for all subgs of current org where the user is a member
+	if search.SuborgRuns == true {
```
frikky, line 28487: What happens if the search is specifically for a single workflow? Do we just remove the filter?

I'm aware we said we would wait with this, but the if statement needs to at least take it into account as to not "randomly" add a bunch of suborg workflows.

frikky (commented):
Nice

shared.go
LalitDeore: Yes, when suborg workflow run search is one and if user select any specific workflow to search than it clear filter for suborg and then only search for that workflow

shared.go
LalitDeore: true, I will update that check.

LalitDeore: Update code in this commit: https://github.com/Shuffle/shuffle-shared/pull/145/commits/42ce12512b411a521b35df04856d246866cbeacc

PR #150, opened 2025-02-18
make cookies http-only, allow to set secure and domain

P4sca1:
Closes https://github.com/Shuffle/Shuffle/issues/1604


frikky:
> Closes [Shuffle/Shuffle#1604](https://github.com/Shuffle/Shuffle/issues/1604)

Cheers! As long as we keep both `session_token` and `__session` this is a great start to refactoring and fixing cookies.

P4sca1:
@frikky This MR Is now ready for review. I cleaned up the code by moving the cookie construction into a function and allowed to change the cookie parameters secure and http-only via env variables.

P4sca1:
@frikky Let's talk about this tomorrow.

frikky:
We will eventually get back to this. We're just in a place where we can't provoke any random bugs right now, and won't have time to test it well enough to merge.

Keeping open until then :))

frikky:
Time to give it a go before the next release :3

This solutions is simple, so shouldn't be too hard to solve some of the potential edgecases.

PR #151, opened 2025-02-24
fixes the current problem with the health API

yashsinghcodes:
Screenshot 2025-02-24 at 9 16 25 PM
Screenshot 2025-02-24 at 9 38 18 PM

When there are more than two health check requests per second, the deleteJunkOpsWorkflow() function sometimes removes the workflow for another requested health check that it uses.

I quickly wrote a fix: if the workflow does not exist or the user is unauthorized, it simply skips the run.

t's still unclear where these requests are coming from, as the Google Script only makes requests at 15-minute intervals.

### DIFF
--- a/health.go
+++ b/health.go
@@ -616,6 +616,13 @@ func RunOpsHealthCheck(resp http.ResponseWriter, request *http.Request) {
 			resp.Write([]byte(`{"success": false, "reason": "High number of requests sent to the backend. Try again later."}`))
 			return
 		}
+
+		if err.Error() == "Unauthorized user saving ops workflow" {
+			log.Printf("[DEBUG] Unauthorized user saving ops workflow. Skipping this run.")
+			resp.WriteHeader(401)
+			resp.Write([]byte(`{"success": false, "reason": "Unauthorized user saving ops workflow."}`))
+			return
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -616,6 +616,13 @@ func RunOpsHealthCheck(resp http.ResponseWriter, request *http.Request) {
 			resp.Write([]byte(`{"success": false, "reason": "High number of requests sent to the backend. Try again later."}`))
 			return
 		}
+
+		if err.Error() == "Unauthorized user saving ops workflow" {
+			log.Printf("[DEBUG] Unauthorized user saving ops workflow. Skipping this run.")
+			resp.WriteHeader(401)
```
frikky, line 622: AUTHORization = 403
AUTHentication = 401

:))

health.go
```diff
@@ -1523,6 +1534,14 @@ func InitOpsWorkflow(apiKey string, OrgId string) (string, error) {
 
 	defer resp.Body.Close()
 
+	// This happend due to deleteJunkOpsWorkflow deleting the workflow before we even save
+	// data. Reason behind is we are making health checks request too fast i.e. less than 
+	// 1s.
+	if resp.StatusCode == 401 {
```
frikky, line 1540: Why is this specialized, and not just together with the != 200 below?

yashsinghcodes: wrote it in a hurry, will fix this in the upcoming PRs.

PR #154, opened 2025-02-28
Fix - app deployment issue

LalitDeore:
@frikky Please review once before merging
1.  Fix specical character in issue in function name
2. Fix same parameter pass two time in same function issue.

### DIFF
--- a/codegen.go
+++ b/codegen.go
@@ -514,20 +514,6 @@ func HandleCheckValidURL(url string) string {
 	return strings.Join(parts, "/")
 }
 
-// Function to remove any special characters from python function name
-func HandleValidatePythonFunctionName(name string) string {
-	// Remove invalid characters (anything except letters, numbers, and underscores)
-	re := regexp.MustCompile(`[^A-Za-z0-9_]`)
-	name = re.ReplaceAllString(name, "_")
-
-	// Ensure it starts with a letter (if not, prefix it with an underscore)
-	if len(name) == 0 || (name[0] < 'A' || (name[0] > 'Z' && name[0] < 'a') || name[0] > 'z') {
-		name = "_" + name
-	}
-
-	return name
-}
-
 // This function generates the python code that's being used.
 // This is really meta when you program it. Handling parameters is hard here.
 func MakePythoncode(swagger *openapi3.Swagger, name, url, method string, parameters, optionalQueries, headers []string, fileField string, api WorkflowApp, handleFile bool) (string, string) {
@@ -739,13 +725,10 @@ func MakePythoncode(swagger *openapi3.Swagger, name, url, method string, paramet
 			}
 		}
 	}
-
 	functionname := strings.ToLower(fmt.Sprintf("%s_%s", method, name))
 	if strings.Contains(strings.ToLower(name), strings.ToLower(method)) {
 		functionname = strings.ToLower(name)
 	}
-
-	functionname = HandleValidatePythonFunctionName(functionname)
 
 	bodyParameter := ""
 	bodyAddin := ""
@@ -2249,6 +2232,13 @@ func FixFunctionName(functionName, actualPath string, lowercase bool) string {
 
 	if lowercase == true {
 		functionName = strings.ToLower(functionName)
+	}
+
+	re := regexp.MustCompile(`[^A-Za-z0-9_]`)
+	functionName = re.ReplaceAllString(functionName, "_")
+
+	if len(functionName) == 0 || (functionName[0] < 'A' || (functionName[0] > 'Z' && functionName[0] < 'a') || functionName[0] > 'z') {
+		functionName = "_" + functionName
 	}
 
 	return functionName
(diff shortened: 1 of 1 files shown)


codegen.go
```diff
@@ -747,14 +747,55 @@ func MakePythoncode(swagger *openapi3.Swagger, name, url, method string, paramet
 	// Extra param for url if it's changeable
 	// Extra param for authentication scheme(s)
 	// The last weird one is the body.. Tabs & spaces sucks.
+
+	// Create a map to track used parameters
+	usedParams := make(map[string]bool)
+
+	addParameter := func(param string) string {
```
0x0elliot, line 754: make this an independent function for cleanliness?

codegen.go
```diff
@@ -2167,6 +2208,10 @@ func FixFunctionName(functionName, actualPath string, lowercase bool) string {
 	functionName = strings.Replace(functionName, " ", "_", -1)
 	functionName = strings.Replace(functionName, "-", "_", -1)
 
+	functionName = strings.Replace(functionName, "%", "", -1)
```
0x0elliot, line 2211: can you try handling this by whitelisting characters instead of blacklisting them? notice how:
-  " " -> "_" and "-" -> "_"
- but, the rest are just stripped. we can't keep dealing with bullshit characters. might as well strip them away and start from a predictable charset (a-z, A-Z, 1-9) 

@frikky any other reviews?

0x0elliot (changes requested):

codegen.go
```diff
@@ -747,14 +747,55 @@ func MakePythoncode(swagger *openapi3.Swagger, name, url, method string, paramet
...
+	// Build the final parameter string avoiding duplicates
 	parsedParameters := fmt.Sprintf("%s%s%s%s%s%s%s",
-		authenticationParameter,
-		urlParameter,
-		fileParameter,
-		parameterData,
-		queryString,
-		bodyParameter,
-		verifyParam,
+		addParameter(strings.TrimPrefix(authenticationParameter, ",")),
```
frikky, line 792: I'm scared of this. Sounds like it would break literally everything.

frikky:
In short: Your code is not at all trustworthy for our problem, unless it's tested across 1000+ apps and doesn't break a single one. I'm scared of deploying it due to it being VERY big changes to the actual function definitions.

This does not feel like a good solution, as it's a solution on the code generation side (python) and not a solution on the control-mechanism of the incoming OpenAPI

LalitDeore:
This two commit ( [commit 1](https://github.com/Shuffle/shuffle-shared/pull/154/commits/703808b7fe6749a288e793a4836db7a05c7e54e0), [commit 2](https://github.com/Shuffle/shuffle-shared/pull/154/commits/5eeae65ccfefe2223f74f5a9cde3d2f58c74e611) ) fixes the following issues:

1. Created a new function to check if the same parameter is passed multiple times to the same function without using a map.
2. Added a function to validate the URL format, ensuring that incorrect URLs like `url}/api/v1/user/{id/new` are corrected to `{url}/api/v1/user/{id}/new`
3. Implemented a regex validation to check that Python function names can only contain letters (a-z, A-Z) and numbers (0-9), with no special characters and replace special character with "_". 

Most app deployments were failing due to these issues. To validate this changes, I deployed 8–10 different apps on Frankfurt, and it worked successfully.

frikky:
As this is related to App Health, I'm assigning it to you @yashsinghcodes. You are testing similar stuff already, and I need you to make sure we can push this code without it causing even a single old app to stop functioning.

Changing names of actions in one place (python) but not another (app.yaml) have a tendency to create chaos.

yashsinghcodes:
Hey @LalitDeore

Found some issue with the function `HandleCheckValidURL` it is creating a problem, when I was testing here is the error I was getting when deploying a new app (cloud function).

Screenshot 2025-03-12 at 12 54 59 AM


Here is the app id of the app I used: a286d101b679672403456f056a4c600f

feel free to let me know if you want any information on it.

frikky:
> Hey @LalitDeore
> 
> Found some issue with the function `HandleCheckValidURL` it is creating a problem, when I was testing here is the error I was getting when deploying a new app (cloud function).
> 
> Screenshot 2025-03-12 at 12 54 59 AM
> Here is the app id of the app I used: a286d101b679672403456f056a4c600f
> 
> feel free to let me know if you want any information on it.

The TWO things I need to see from this PR to be confident in merging it:
1. There NEEDS to be a sync between the Python function name and the Api.yaml function. Right now, you are ONLY fixing the actual Python code, but leaving out the user experience entirely.  
2. It needs to solve literally 100% of the current random code generator issues we have. Right now, I can still see a ton of cloud functions that have errors. If this PR works, they should not exist anymore.

frikky:
Closing as you haven't sent any feedback in a while. 

This kind of PR is VERY scary to merge without proof of it not breaking app deployments.

PR #156, opened 2025-03-03
check if health check already exist in the queue

yashsinghcodes:


health.go
```diff
@@ -429,6 +429,51 @@ func deleteJunkOpsWorkflow(ctx context.Context, workflowHealth WorkflowHealth) e
 	return nil
 }
 
+func checkQueueForHealthRun(ctx context.Context) error{
+	baseUrl := "https://shuffler.io"
+	if os.Getenv("SHUFFLE_CLOUDRUN_URL") != "" {
```
frikky, line 434: Just always check BASE_URL > CLOUDRUN_URL, as CLOUDRUN_URL is the override we use everywhere

health.go
```diff
@@ -429,6 +429,51 @@ func deleteJunkOpsWorkflow(ctx context.Context, workflowHealth WorkflowHealth) e
...
+		log.Printf("[DEBUG] Setting the baseUrl for health check to %s", baseUrl)
+		baseUrl = os.Getenv("SHUFFLE_CLOUDRUN_URL")
+	}
+
+	if project.Environment == "onprem" {
```
frikky, line 439: This may be wrong, so instead of `== "onprem"` do `!= "cloud"`

health.go
```diff
@@ -429,6 +429,51 @@ func deleteJunkOpsWorkflow(ctx context.Context, workflowHealth WorkflowHealth) e
...
+		log.Printf("[DEBUG] Onprem environment. Setting base url to localhost: for delete")
+		baseUrl = "http://localhost:5001"
+		if os.Getenv("BASE_URL") != "" {
+			baseUrl = os.Getenv("BASE_URL")
+		}
+	}
+
+
+	client := GetExternalClient(baseUrl)
+	fullUrl := fmt.Sprintf("%s/api/v1/workflows/queue", baseUrl)
```
frikky, line 449: Can't you read this directly from the database? Using the API seems silly lol.

health.go
```diff
@@ -565,6 +610,41 @@ func RunOpsHealthCheck(resp http.ResponseWriter, request *http.Request) {
 		}
 	} else {
 		// FIXME: Add a check for if it's been <interval> length at least between runs. This is 15 minutes by default.
+		err := checkQueueForHealthRun(ctx)
+		if err != nil {
+			log.Printf("[ERROR] Failed running health check (4): %s", err)
```
frikky, line 615: Shouldn't this then fail out? Why are we not handling the error?

frikky (commented):
:)

yashsinghcodes:
ready for round 2 of review @frikky :)

frikky:
I guess we run with it. Did this fix the onprem "admin" issue?

PR #162, opened 2025-04-03
removed github backup info from the json file

yashsinghcodes:
I think we should, also add some kind of warning in the frontend - informing user that there workflow could contain sensitive data and pushing them to public repo can create problems.

We can remove sensitive information from the action, but then it won't be a proper backup.


yashsinghcodes:
Hey @frikky I have used go-routine function for some of the non-returing function calls. This had improved the performance when testing locally.

frikky:
> Hey @frikky I have used go-routine function for some of the non-returing function calls. This had improved the performance when testing locally.

How can you assure that the context doesn't time out or finish before it's used in the goroutine?

PR #163, opened 2025-04-07
Fix - increase app and workflow run limit from parent org to suborg

LalitDeore:
https://github.com/Shuffle/shaffuru/pull/533

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -11867,7 +11867,7 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
 	}
 
 	admin := false
-	if tmpData.OrgId != user.ActiveOrg.Id || fileId != user.ActiveOrg.Id {
+	if (tmpData.OrgId != user.ActiveOrg.Id || fileId != user.ActiveOrg.Id) && !tmpData.SyncFeatures.Editing {
 		log.Printf("[WARNING] User can't edit org %s (active: %s)", fileId, user.ActiveOrg.Id)
 		if !user.SupportAccess {
 			resp.WriteHeader(401)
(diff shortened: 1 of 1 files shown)


shared.go
```diff
@@ -11867,7 +11867,7 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
 	}
 
 	admin := false
-	if tmpData.OrgId != user.ActiveOrg.Id || fileId != user.ActiveOrg.Id {
+	if (tmpData.OrgId != user.ActiveOrg.Id || fileId != user.ActiveOrg.Id) && !tmpData.SyncFeatures.Editing {
```
frikky, line 11870: Wait, what is this Editing part? I'm assuming it's related to a parent org admin controlling a child org, but I'm unsure how it works. 

Is it sent in with the request (tmpData) or stored as an entry somehow? If it's just sent in; how are we making sure the user should be able to Edit?

frikky (commented):
I'm just confused about the access rights of editing another org, as only support users used to be able to change the features in the past.

Are we preventing them from changing their own org? Can "any" admin change it?

I think we need to lock this down a bit more.

shared.go
LalitDeore: tmpData is the org data sent in the request body. When `tmpData.SyncFeature.Editing` is true, it indicates that the user is editing the sync features. In this case, we should skip the usual check, because while editing a sub-organization, the suborg.id doesnt match with the user's active organization. That's why this condition needs to be bypassed during sync feature editing. This condition doesn't impact while editing the current org.

Additionally, I’ve added a validation step to ensure that the user's active organization actually contains the sub-organization being edited. This is to make sure the user isn’t trying to edit someone else’s organization.

PR #173, opened 2025-05-05
Added geoapify on backend

Monilprajapati:
- Sending the user region in response when the user is logged in to initialise the GA based on location


frikky (commented):
Optimize

frikky:
@Monilprajapati still waiting for feedback as you said this is a blocker (:

PR #174, opened 2025-05-07
fix: setcache keep the distribution state, and health workflow don't create org notifications

yashsinghcodes:
…create org notification


shared.go
```diff
@@ -18377,7 +18377,15 @@ func HandleSetCacheKey(resp http.ResponseWriter, request *http.Request) {
 		tmpData.Category = ""
 	}
 
+
 	tmpData.Key = strings.Trim(tmpData.Key, " ")
+	
+	// Check if cache already existed and if distributed
+	cacheData, err := GetCacheKey(ctx, tmpData.Key, tmpData.Category)
```
frikky, line 18384: Wait, what is this?

Also "GetCache()" vs "GetCacheKey()" is confusing D:

yashsinghcodes: Oh yes this was the fix for 

"If we change the content of a datastore entry, then the distribution to the suborgs stops. That means that the check (“Distribution YES/NO) gets removed. Especially then, when using the api to access this datastore, the distribution stops, which is breaking our workflows in the suborgs."

this problem where, the datastore entry looses it's distribution status when updated.

PR #179, opened 2025-05-19
Keep context of the distributed workflow env

yashsinghcodes:
If the parent workflow changes its environment to something that is not distributed, we don't want to update the distributed workflow, as this would result in the distributed workflow incorrectly having 'cloud' as its environment.


yashsinghcodes:
Hello @frikky, added some comment ready for another review :)))

PR #182, opened 2025-05-20
change theme as user change

LalitDeore:


frikky:
Why is this a standalone API and not just a user edit?

LalitDeore:
> Why is this a standalone API and not just a user edit?

I have removed new api and added theme change in existing api.

PR #185, opened 2025-05-30
Partner and usecase system

Monilprajapati:
- CloudSync: Created two new indexes on Algolia for partners and use cases to make it easier to display those pages on the on-prem side. Also created their upload functions and the partnerSearch function and usecaseDeletion function from algolia.
- db-connector: Created setter and getter functions for partners and usecases.
- shared.go: Added functions getPartner, publishPartner, getUsecases, and publishUsecases, deleteUsecase.
- struct.go: Added new structs for partner and use case for both DB and Algolia search.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -25546,12 +25546,12 @@ func HandleGetPartnerUsecases(resp http.ResponseWriter, request *http.Request) {
 
 	ctx := GetContext(request)
 
-	// Check if the partner is public or not
-	partner, err := HandleAlgoliaPartnerSearch(ctx, Id)
+	partner, err := GetPartnerById(ctx, Id)
+
 	if err != nil {
-		log.Printf("[WARNING] Partner with Id %s is not public: %s", Id, err)
-		resp.WriteHeader(401)
-		resp.Write([]byte(`{"success": false}`))
+		log.Printf("[ERROR] Failed to get partner: %v", err)
+		resp.WriteHeader(http.StatusInternalServerError)
+		resp.Write([]byte(`{"success": false, "reason": "Failed to get partner"}`))
 		return
 	}
 
@@ -25567,7 +25567,7 @@ func HandleGetPartnerUsecases(resp http.ResponseWriter, request *http.Request) {
 	}
 
 	// Filter to only include public usecases
-	if partner.ObjectID != user.ActiveOrg.Id {
+	if partner.Id != user.ActiveOrg.Id {
 		for _, usecase := range allUsecases {
 			if usecase.Public {
 				usecases = append(usecases, usecase)
@@ -25766,7 +25766,7 @@ func HandleDeleteUsecase(resp http.ResponseWriter, request *http.Request) {
 
 	// Delete usecase from database
 	nameKey := "Usecases"
-	DeleteCache(ctx, fmt.Sprintf("%s_%s", nameKey, usecase.CompanyInfo.Id))
+	DeleteCache(ctx, fmt.Sprintf("%s_partner_%s", nameKey, usecase.CompanyInfo.Id))
 	err = DeleteKey(ctx, nameKey, usecase.Id)
 	if err != nil {
 		log.Printf("[WARNING] Failed deleting usecase: %v", err)


frikky (commented):
Look through EVERY function for broken access control.

It shows up again and again, and I think with this code, anyone can steal anyones Partner control AND their Usecases. 

Learn from this, and ALWAYS keep worst case scenarios in mind when writing code. People do bad shit on purpose.

cloudSync.go
```diff
@@ -547,6 +605,117 @@ func HandleAlgoliaCreatorDeletion(ctx context.Context, userId string) (error) {
...
+			PartnerName:        usecase.CompanyInfo.Name,
+			PartnerId:          usecase.CompanyInfo.Id,
+			Name:               usecase.MainContent.Title,
+			Description:        usecase.MainContent.Description,
+			Categories:         usecase.MainContent.Categories,
+			SourceAppType:      usecase.MainContent.SourceAppType,
+			DestinationAppType: usecase.MainContent.DestinationAppType,
+			PublicWorkflowID:   usecase.MainContent.PublicWorkflowID,
+			TimeEdited:         timeNow,
+			Published: 		    usecase.Published,
```
frikky, line 658: Why is this a field in Algolia? Everything in Algolia should be default be public, so this is redundant. 

PS: Other places, we have called publishing "Public", not "Published". Please follow convention

structs.go
```diff
@@ -925,6 +926,60 @@ type LeadInfo struct {
...
 
+// Partners Structs
+type PartnerType struct {
+	TechPartner         bool `json:"tech_partner,omitempty" datastore:"tech_partner"`
+	IntegrationPartner  bool `json:"integration_partner,omitempty" datastore:"integration_partner"`
+	DistributionPartner bool `json:"distribution_partner,omitempty" datastore:"distribution_partner"`
+	ServicePartner      bool `json:"service_partner,omitempty" datastore:"service_partner"`
+}
+
+type Partner struct {
```
frikky, line 937: Why is there nothing in here about it being Public or not?

structs.go
```diff
@@ -925,6 +926,60 @@ type LeadInfo struct {
...
+		SourceAppType      string   `datastore:"sourceAppType" json:"sourceAppType"`
+		DestinationAppType string   `datastore:"destinationAppType" json:"destinationAppType"`
+	} `datastore:"mainContent" json:"mainContent"`
+	Navigation struct {
+		Items []struct {
+			Name    string   `datastore:"name" json:"name"`
+			Content []string `datastore:"content" json:"content"`
+		} `datastore:"items" json:"items"`
+	} `datastore:"navigation" json:"navigation"`
+	Published bool  `datastore:"published" json:"published"`
```
frikky, line 978: Call it "Public" not "Published" to follow convention from other places

structs.go
```diff
@@ -1734,6 +1789,34 @@ type AlgoliaSearchCreator struct {
...
+	TimeEdited  int64    `json:"time_edited"`
+	SquareImage string   `json:"square_image"`
+	Name        string   `json:"name"`
+	Description string   `json:"description"`
+	PartnerType []string `json:"partner_type"`
+	Solutions   []string `json:"solutions"`
+	Country     string   `json:"country"`
+	Region      string   `json:"region"`
+	OrgId       string   `json:"org_id"`
+	IsPublished bool     `json:"ispublished"`
```
frikky, line 1803: Why is it here and not in the real Partner struct lol

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+	}
+
+	if len(tmpData.CompanyInfo.Id) == 0 {
+		log.Printf("[WARNING] No partner ID provided for usecase %s", tmpData.CompanyInfo.Name)
+		resp.WriteHeader(400)
+		resp.Write([]byte(`{"success": false, "reason": "No company ID provided"}`))
+		return
+	}
+
+	partner, err := GetPartnerById(ctx, tmpData.CompanyInfo.Id)
```
frikky, line 25420: If you just put `user.ActiveOrg.Id` instead of `tmpData.CompanyInfo.Id` then the next if-statement is redundant

shared.go
```diff
@@ -25012,6 +25271,352 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+	if len(tmpData.Id) == 0 {
+		tmpData.Id = uuid.NewV4().String()
+	} else if tmpData.Id != Id && len(Id) > 0 {
+		// Handle mismatch between URL ID and body ID
+		log.Printf("[WARNING] ID mismatch: URL ID %s vs Body ID %s", Id, tmpData.Id)
+		resp.WriteHeader(http.StatusBadRequest)
+		resp.Write([]byte(`{"success": false, "reason": "ID mismatch between URL and request body"}`))
+		return
+	}
+
```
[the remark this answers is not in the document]
frikky: I still don't see anywhere that it checks who owns the usecase. This should be a forceful check - not whether Partner ID and Active Org are the same, but whether Active orgId and Usecase orgId are the same.

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+	_, err = HandleAlgoliaUsecaseUpload(ctx, tmpData, overwrite)
+	if err != nil {
+		log.Printf("[WARNING] Failed publishing usecase to Algolia: %v", err)
+	}
+
+	err = SetUsecaseNew(ctx, &tmpData)
+	if err != nil {
+		log.Printf("[WARNING] Failed publishing usecase: %v", err)
```
frikky, line 25458: This should be an [ERROR] as we want to know if this _ever_ occurs

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+		// Handle mismatch between URL ID and body ID
+		log.Printf("[WARNING] ID mismatch: URL ID %s vs Body ID %s", Id, tmpData.Id)
+		resp.WriteHeader(http.StatusBadRequest)
+		resp.Write([]byte(`{"success": false, "reason": "ID mismatch between URL and request body"}`))
+		return
+	}
+
+	_, err = HandleAlgoliaUsecaseUpload(ctx, tmpData, overwrite)
+	if err != nil {
+		log.Printf("[WARNING] Failed publishing usecase to Algolia: %v", err)
```
frikky, line 25453: This should be an [ERROR] as we need to know if this occurs

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+
+	overwrite := false
+
+	if len(tmpData.Id) > 0 {
+		overwrite = true
+	}
+
+	if len(tmpData.Id) == 0 {
+		tmpData.Id = uuid.NewV4().String()
+	} else if tmpData.Id != Id && len(Id) > 0 {
```
frikky, line 25443: There is nowhere that you are validating the actual usecaseId for some reason? If it's an edit, you NEED to validate that the owner of that ID is the current usecase. Otherwise orgs can overwrite eachothers' Usecases

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+		resp.Write([]byte(`{"success": false}`))
+		return
+	}
+
+	if len(partner.Usecases) == 0 {
+		partner.Usecases = []string{}
+	}
+
+	// Check if the usecase is already in the partner's usecases
+	found := false
```
frikky, line 25469: You should NOT need to do this. Usecase query should be its own API, where you filter by the Partner ID - it should not need to be tracked within the partner itself.

Just imagine it having 1000+ usecases, and having a huge blob within the partner struct. This is not really a big deal, but is REALLY not ideal development-wise

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+		_, err := HandleAlgoliaPartnerSearch(ctx, Id)
+		if err != nil {
+			log.Printf("[WARNING] Partner with Id %s is not public: %s", Id, err)
+			resp.WriteHeader(401)
+			resp.Write([]byte(`{"success": false}`))
+			return
+		}
+
+		// Gettting the partner's usecases
+		usecases, err = GetPartnerUsecases(ctx, Id)
```
frikky, line 25539: No error validation..?

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+
+	if err != nil {
+		log.Printf("[ERROR] Failed to get usecases: %v", err)
+		resp.WriteHeader(http.StatusInternalServerError)
+		resp.Write([]byte(`{"success": false, "reason": "Failed to get usecases"}`))
+		return
+	}
+
+	// Marshal the response
+	response, err := json.Marshal(struct {
```
frikky, line 25557: Pre-make a struct - don't do these in-memory shenanigans

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+			log.Printf("[WARNING] Partner with Id %s is not public: %s", Id, err)
+			resp.WriteHeader(401)
+			resp.Write([]byte(`{"success": false}`))
+			return
+		}
+
+		// Gettting the partner's usecases
+		usecases, err = GetPartnerUsecases(ctx, Id)
+	} else {
+		var usecase UsecaseInfo
```
frikky, line 25541: Why isn't this a separate API lol

In CRUD requests, you typically separate Listing and Get because maintaining the code becomes WAAAAY easier

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+	}
+
+	if len(usecase.CompanyInfo.Id) == 0 {
+		log.Printf("[WARNING] No partner ID provided for usecase %s", usecase.CompanyInfo.Name)
+		resp.WriteHeader(400)
+		resp.Write([]byte(`{"success": false, "reason": "No company ID provided"}`))
+		return
+	}
+
+	partner, err := GetPartnerById(ctx, usecase.CompanyInfo.Id)
```
frikky, line 25642: Can't you just validate if user.ActiveOrg.Id == usecase.CompanyInfo.Id here? Do we even need the partner check? 

Less api calls \o/

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+		log.Printf("[AUDIT] User %s (%s) tried to delete usecase %s from partner %s (%s) but doesn't have access", user.Username, user.Id, usecase.Id, partner.Name, partner.Id)
+		resp.WriteHeader(403)
+		resp.Write([]byte(`{"success": false, "reason": "Unauthorized access to partner's usecases"}`))
+		return
+	}
+
+	// Check if the usecase is in the partner's usecases
+	found := false
+
+	for i, usecaseId := range partner.Usecases {
```
frikky, line 25660: I still just think we shouldn't track usecases in partner.Usecases at all. Just always search for them in the database properly based on partnerId/orgId.

shared.go
```diff
@@ -25219,6 +25341,374 @@ func HandleGetUsecase(resp http.ResponseWriter, request *http.Request) {
...
+		if usecaseId == usecase.Id {
+			found = true
+			// Remove the usecase from the partner's usecases
+			partner.Usecases = append(partner.Usecases[:i], partner.Usecases[i+1:]...)
```
frikky, line 25664: This kind of shit can and will always break as you aren't doing proper bounds-checking. We aren't constrained by this kind of CPU issue, so don't try to optimise minimal things like this that don't need optimising.

The trick: Make a new array and append instead.

db-connector.go
```diff
@@ -12674,6 +12852,145 @@ func GetUsecase(ctx context.Context, name string) (*Usecase, error) {
...
+                err = json.Unmarshal(cacheBytes, &cachedUsecases)
+                if err == nil {
+                    return cachedUsecases, nil
+                }
+            }
+        }
+    }
+
+    // Get from datastore if not in cache
+    q := datastore.NewQuery(nameKey).Filter("companyInfo.id=", partnerId)
```
frikky, line 12971: Since we are querying like this, there is no need to track ID's within []partner.Usecases{} as a string at all. It's never used, and is redundant for no reason.

shared.go
```diff
@@ -9105,6 +9105,128 @@ func HandleGetUsers(resp http.ResponseWriter, request *http.Request) {
...
+	partners, err := GetAllPartners(ctx)
+	if err != nil {
+		log.Printf("[ERROR] Failed to get partners: %v", err)
+		resp.WriteHeader(http.StatusInternalServerError)
+		resp.Write([]byte(`{"success": false, "reason": "Failed to get partners"}`))
+		return
+	}
+
+	// Marshal the response
+	response, err := json.Marshal(struct {
```
frikky, line 9134: Pre-make structs you want to use. Don't do this.

shared.go
```diff
@@ -9105,6 +9105,128 @@ func HandleGetUsers(resp http.ResponseWriter, request *http.Request) {
...
+		}
+	}
+
+	ctx := GetContext(request)
+	partners, err := GetAllPartners(ctx)
+	if err != nil {
+		log.Printf("[ERROR] Failed to get partners: %v", err)
+		resp.WriteHeader(http.StatusInternalServerError)
+		resp.Write([]byte(`{"success": false, "reason": "Failed to get partners"}`))
+		return
```
frikky, line 9130: There is no validation anywhere for whether a partner is public or not????

shared.go
```diff
@@ -9105,6 +9105,128 @@ func HandleGetUsers(resp http.ResponseWriter, request *http.Request) {
...
+	var err error
+
+	if len(partnerId) == 0 {
+		log.Printf("[ERROR] Partner ID is missing in request: %s", request.URL.String())
+		resp.WriteHeader(http.StatusBadRequest)
+		resp.Write([]byte(`{"success": false, "reason": "Missing partner ID"}`))
+		return
+	}
+
+	partner, err = GetPartnerById(ctx, partnerId)
```
frikky, line 9188: Where is the validation for whether the partner is public or the user has access to it??

frikky (commented):
The lesson definitely hasn't been learned here. Please fix all this

shared.go
Monilprajapati: Done

shared.go
Monilprajapati: Done

shared.go
Monilprajapati: True.
Removed the usecases array from the Partner struct since we're now fetching partner usecases directly through a DB query.

shared.go
Monilprajapati: It was below both API calls earlier, but I’ve separated those two functions now.

shared.go
Monilprajapati: Done

structs.go
Monilprajapati: It is there on line 953.

PR #186, opened 2025-06-02
[feature] - skip sso fro the admin on suborg

LalitDeore:


shared.go
```diff
@@ -11107,39 +11107,66 @@ func HandleChangeUserOrg(resp http.ResponseWriter, request *http.Request) {
...
+				log.Printf("[ERROR] Failed getting parent org %s for suborg %s: %s", org.CreatorOrg, org.Id, err)
+				resp.WriteHeader(401)
+				resp.Write([]byte(`{"success": false, "reason": "Failed getting parent org for suborg"}`))
+				return
+			}
 
-			baseSSOUrl = GetOpenIdUrl(request, *org)
+			if parentOrg.SSOConfig.SkipSSOForAdmins {
+				for _, orgUser := range parentOrg.Users {
+					if orgUser.Id == user.Id && orgUser.Role == "admin" {
```
frikky, line 11124: Nitpicking:

- When using if statements to find a single value within a loop, ALWAYS break

This isn't super important in this specific case due to the size of users not being in thousands, but it's a general good rule to follow

shared.go
```diff
@@ -11107,39 +11107,66 @@ func HandleChangeUserOrg(resp http.ResponseWriter, request *http.Request) {
...
+			}
 		}
 
-		if !strings.HasPrefix(baseSSOUrl, "http") {
-			log.Printf("[ERROR] SSO URL for %s (%s) is invalid: %s", org.Name, org.Id, baseSSOUrl)
-			//resp.WriteHeader(401)
-			//resp.Write([]byte(`{"success": false, "reason": "SSO URL is invalid"}`))
-			//return
+		if skipSSO {
+			log.Printf("[DEBUG] User %s (%s) is skipping SSO for suborg %s (%s)", user.Username, user.Id, org.Name, org.Id)
```
frikky, line 11134: I think this should be using [AUDIT] as it's an audit log for the user getting access to something

shared.go
```diff
@@ -11107,39 +11107,66 @@ func HandleChangeUserOrg(resp http.ResponseWriter, request *http.Request) {
...
+				baseSSOUrl = GetOpenIdUrl(request, *org)
 			}
 
-			b, err := json.Marshal(ssoResponse)
-			if err != nil {
-				log.Printf("[ERROR] Failed marshalling SSO response: %s", err)
-				resp.Write([]byte(`{"success": false}`))
+			if !strings.HasPrefix(baseSSOUrl, "http") {
+				log.Printf("[ERROR] SSO URL for %s (%s) is invalid: %s", org.Name, org.Id, baseSSOUrl)
+				//resp.WriteHeader(401)
```
frikky, line 11147: Any reason this is commented out?

shared.go
```diff
@@ -11831,21 +11859,11 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
 
 	//Update mfa required value
 	if tmpData.MFARequired != org.MFARequired {
+		log.Printf("[DEBUG] Setting MFA required to %t for org %s (%s)", tmpData.MFARequired, org.Name, org.Id)
```
frikky, line 11862: Small recent change:

[DEBUG] will be removed from our logs unless DEBUG=true environment variable is set.

What this means is that if you want the logs to persist at all times, use [INFO]

DEBUG
INFO
WARNING
ERROR

[AUDIT] for access right logs

frikky (commented):
Very minor

shared.go
LalitDeore: It was already commented out, not sure why it was commeted out. I have uncomment it now.

PR #201, opened 2025-06-24
improve the suborg delete api

LalitDeore:


frikky:
A question that shows up here: Should it be possible to delete parent orgs? They are the same, just without a reference.

PR #204, opened 2025-07-04
[draft]: Public workflow fix

Monilprajapati:
- Will write explanation soon in this PR.


frikky:
If it's a draft, make it a draft :P

frikky:
I assume this will be fixed with the others, or is it separate?

frikky:
> > I assume this will be fixed with the others, or is it separate?
> 
> Could you please clarify what you meant by that? Just want to make sure I understand correctly.

You said you would make a new PR with the changes we discussed so that we can move on :P

frikky:
Merging this even though it's not exactly going to make things easy to understand in the future..

PR #206, opened 2025-07-13
Added email scope for openid authentication url

LalitDeore:


frikky:
> > Uhm, why is this necessary? I'm quite sure this will break something.
> > Provide some real version of it working and failing please.
> 
> The issue was that some OIDC SSO providers, such as Google and Authentik, do not include the user's email ID in the response by default. To receive the email ID, the email scope must be added to the request URL. I tested this with both Authentik and Keycloak, and it works correctly in both cases.
> 
> https://www.loom.com/share/21521b02157d4f76b9fc3529a3827434

Please try a few more vendors as well. Any default added scope will for sure cause some kind of service disruption, unless it is part of the OIDC standard for exactly our type of auth

frikky:
> > > > Uhm, why is this necessary? I'm quite sure this will break something.
> > > > Provide some real version of it working and failing please.
> > > 
> > > 
> > > The issue was that some OIDC SSO providers, such as Google and Authentik, do not include the user's email ID in the response by default. To receive the email ID, the email scope must be added to the request URL. I tested this with both Authentik and Keycloak, and it works correctly in both cases.
> > > https://www.loom.com/share/21521b02157d4f76b9fc3529a3827434
> > 
> > 
> > Please try a few more vendors as well. Any default added scope will for sure cause some kind of service disruption, unless it is part of the OIDC standard for exactly our type of auth
> 
> I also tested this with Azure, and it worked as expected. According to the [OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html), the `openid `scope is required for authentication, while additional scopes like `profile `and `email `can be included to request access to the user's profile and email information.

Alright, I'll trust you on this then.

Scope is not my problem. What you put in the scope is. I've had too many issues with Oauth2/OpenID scopes in the past and this just feels like it will break... something.

PR #218, opened 2025-08-07
[feature] - send sales mail

LalitDeore:


shared.go
```diff
@@ -12542,6 +12543,211 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
...
+		localPart := parts[0]
+		domainPart := parts[1]
+
+		domainLower := strings.ToLower(domainPart)
+
+		if strings.Contains(domainLower, "gmail.com") ||
+			strings.Contains(domainLower, "yahoo.com") ||
+			strings.Contains(domainLower, "aol.com") ||
+			strings.Contains(domainLower, "outlook.com") ||
+			strings.Contains(domainLower, "infopercept.com") ||
```
frikky, line 12620: Wtf is this doing in here lol

And what is this list for? I don't get it

shared.go
```diff
@@ -12542,6 +12543,211 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
...
+				orgName = string(runes)
+			}
+		}
+	} else {
+		orgName = org.Name
+	}
+
+	subject := fmt.Sprintf("Shuffle and %s", orgName)
+
+	frameworkItems := []string{
```
frikky, line 12643: Why are we just using these five?

shared.go
```diff
@@ -12542,6 +12543,211 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
...
+		yourApps += "\n"
+	}
+
+	if len(usecases) > 5 && !strings.HasSuffix(usecases, "\n") {
+		usecases += "\n"
+	}
+
+	personalizedMessage := `Hello,
+
+I noticed you've been exploring Shuffle and setting up some workflows - that's awesome! It seems like you might not be getting everything you’re looking for just yet, and I’d love to help. Would you be open to a quick chat?
```
frikky, line 12714: Why isn't this an LLM that's personalising it EVERY TIME instead of this shitty thing? That way we can actually personalise properly.

Also, what is the point of this being an API when it should just be automatic in the background?

frikky (commented):
:))

LalitDeore:
Closing this PR. As we will handle email automation through workflows instead of the API. @ayush0033

PR #220, opened 2025-08-18
AI improvements: on-prem detection, and cloud usage limits

satti-hari-krishna-reddy:
## Things added 

* Added a new `RunAiQueryWithModel` function to allow overriding the AI model per request, and refactored `RunAiQuery` to use this for more flexible model selection. Model fallback logic distinguishes between cloud and on-prem environments

* Extracted the complex system prompt for workflow breakdown into a new helper function (`getTaskBreakdown`), improving code clarity

* Added AI metadata to workflow struct for identifying AI-generated workflows.
* Added `CheckAndUpdateAIAvailability` function to check for availability of self-hosted/local models.

* AI usage limits enforced only for Cloud usage (skipped for self-hosted/local models).

**Bug Fixes**
* Fixed condition assignment in workflow branches to correctly update the referenced branch by index, preventing missed condition attachments.


ai.go
```diff
@@ -7123,10 +7123,20 @@ func GenerateSingulWorkflows(resp http.ResponseWriter, request *http.Request) {
...
+	return RunAiQueryWithModel(systemMessage, userMessage, "", incomingRequest...)
+}
+
+func RunAiQueryWithModel(systemMessage, userMessage, modelOverride string, incomingRequest ...openai.ChatCompletionRequest) (string, error) {
 	cnt := 0
 	maxTokens := 5000
 	maxCharacters := 100000
 
+	// Use override model if provided, otherwise use the global model
+	selectedModel := model
```
frikky, line 7135: Wtf is this? Why can't you just supply incomingRequest? Why would we want to maintain two extremely similar functions? 

All lines of code are liability. Please don't add this when it could be done without the old function. Remove one of the functions.

ai.go
```diff
@@ -7123,10 +7123,20 @@ func GenerateSingulWorkflows(resp http.ResponseWriter, request *http.Request) {
...
+	if modelOverride != "" {
```
frikky, line 7136: This is silly

ai.go
```diff
@@ -7166,13 +7176,13 @@ func RunAiQuery(systemMessage, userMessage string, incomingRequest ...openai.Cha
...
 	chatCompletion := openai.ChatCompletionRequest{
-		Model: model,
+		Model: selectedModel,
 		Messages: []openai.ChatCompletionMessage{},
 		MaxTokens:   maxTokens,
 	}
 
 	// Newer models (GPT-4o, o1, gpt-5 etc.) require max_completion_tokens instead of max_tokens
-	if strings.HasPrefix(model, "gpt-4o") || strings.HasPrefix(model, "o1") || strings.HasPrefix(model, "gpt-5") || model == "o4-mini" {
+	if strings.HasPrefix(selectedModel, "gpt-4o") || strings.HasPrefix(selectedModel, "o1") || strings.HasPrefix(selectedModel, "gpt-5") || selectedModel == "o4-mini" {
```
frikky, line 7185: This whole thing is a mess and shouldn't necessarily be in this function at all

shared.go
```diff
@@ -3435,6 +3435,89 @@ func HandleApiAuthentication(resp http.ResponseWriter, request *http.Request) (U
 	return User{}, errors.New("Missing authentication")
 }
 
+func CheckAndUpdateAIAvailability(ctx context.Context, orgId string) {
```
frikky, line 3438: Lol this is silly - especially when it's not used in realtime. 

Things can go up and down, and any and all status checks have to be realtime at all times.

shared.go
```diff
@@ -3435,6 +3435,89 @@ func HandleApiAuthentication(resp http.ResponseWriter, request *http.Request) (U
...
+	client := http.Client{
+		Timeout: 3 * time.Second,
+	}
+	
+	headUrl := aiRequestUrl
+	if strings.HasSuffix(headUrl, "/v1") {
+		headUrl = strings.TrimSuffix(headUrl, "/v1")
+	}
+	
+	resp, err := client.Head(headUrl)
```
frikky, line 3485: Are you 100% sure ALL relevant models support HEAD requests? And that their url is ALWAYS /v1?

Actually testing the integration can be done at runtime otherwise. When they actually want something, you reply with the direct response from the vendor - not with a made-up check like this (at least that's what it looks like)

shared.go
```diff
@@ -3435,6 +3435,89 @@ func HandleApiAuthentication(resp http.ResponseWriter, request *http.Request) (U
...
+		log.Printf("[DEBUG] OPENAI_API_URL not set, AI not available for on-prem")
+		return false
+	}
+
+	apiKey := os.Getenv("OPENAI_API_KEY")
+	if len(apiKey) == 0 {
+		log.Printf("[DEBUG] OPENAI_API_KEY not set, Trying to connect without API key")
+	}
+
+	selectedModel := os.Getenv("AI_MODEL")
```
frikky, line 3469: Shouldn't we have a "standard" model for different tools? May take a lot of time tho 🤔

frikky (commented):
Heyo! I'm merging, but solve these things either way. The test mechanisms are heavily over-engineered (without proper tests..) instead of just being direct environment variable checks onprem, which they should be.

PR #225, opened 2025-08-26
Fix: Long SSO certificates not being added due to indexed string size limit

Monilprajapati:
- The issue was that some SSO certificates were too large and couldn’t be stored the db in the indexed field (`SSOCertificate`).
- I didn’t want to make the existing `SSOCertificate` non-indexed, because we still rely on it for lookups of short certs.
- Also couldn’t switch everything to just a hash, since older orgs don’t have one and that would break backward compatibility.
- So the approach I took was:
  - Keep short certs in the existing `SSOCertificate` field (indexed, works as before).
  - For long certs, store them in a new field `SSOLongCertificate` (noindex), and also keep a hash in `SSOCertificateHash` for lookups.
- On save:
  - If the cert is short → store it as usual.
  - If it’s long → move it to `SSOLongCertificate`, save the hash, and clear the indexed field.
- On read:
  - First try to find a match with the short cert.
  - If that fails, fall back to the hash and check against the long cert.
- This way, old orgs continue to work exactly as before, while new/long certs are supported without breaking anything.

Here’s the frontend [PR](https://github.com/Shuffle/shaffuru/pull/656) for it.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -12220,10 +12220,21 @@ func HandleEditOrg(resp http.ResponseWriter, request *http.Request) {
 
 	if len(tmpData.SSOConfig.SSOCertificate) > 0 {
 		savedCert := fixCertificate(tmpData.SSOConfig.SSOCertificate)
-
+		certLen := len([]byte(savedCert))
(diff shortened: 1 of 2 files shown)


frikky:
Lol ok so it's indexing based 🤔

PR #228, opened 2025-08-26
Refactor user message construction for multimodal chat completions

satti-hari-krishna-reddy:
## Changes made

* Dynamically build MultiContent for user messages, including text only if present when generating workflows using images


satti-hari-krishna-reddy:
@frikky 👍? or 👎

PR #337, opened 2026-02-26
fix: normalize environment casing in workflow actions and triggers

0x0elliot:


frikky:
Closing it lol. I guess it's a problem we need to look into and solve tho

PR #343, opened 2026-03-12
Lot's of agent stuff again

satti-hari-krishna-reddy:
Here are the key changes:

- Changed abort status from `"FINISHED"` to `"ABORTED"` throughout the codebase for clearer execution state tracking
- Moved abort logic into the `abortAgentExecution()` function, replacing repetitive error handling code in ~7 locations
- Fixed Action Reference bug where the wrong action object was being passed to the agent continuation logic. Previously, it was using actionResult.Action, but now it properly retrieves the original action from the execution results array, ensuring the agent has the correct state and context when continuing execution. (Look at line 17846 in `shared.go`)
- Added context (`ctx`) as a parameter to `HandleAiAgentExecutionStart()` 
- Added validation for caller function info with trace ID logging for better debugging
- Added `"http"` into the list of available apps for agent decisions), making the agent work with HTTP functionality again.

- Added `AI_AGENT_LLM_FAILURE` log prefix for all LLM-related errors
- Includes detailed error context: org ID, status codes, error types, and raw responses
- Helps distinguish between different failure modes (auth errors, rate limits, parsing errors, etc.)
- Added caller function and trace ID logging for agent starts
- Better separation between `ABORTED`, `FAILURE`, and `FINISHED` states

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -189,7 +189,7 @@ func HandleCors(resp http.ResponseWriter, request *http.Request) bool {
 	}
 
 	//resp.Header().Set("Access-Control-Allow-Origin", "http://localhost:8000")
-	resp.Header().Set("Access-Control-Allow-Headers", "Content-Type, Accept, X-Requested-With, remember-me, Org-Id, Org, Authorization, X-Debug-Url")
+	resp.Header().Set("Access-Control-Allow-Headers", "Content-Type, Accept, X-Requested-With, remember-me, Org-Id, Org, Authorization, X-Debug-Url, X-Internal-Caller, X-Trace-ID")
 	resp.Header().Set("Access-Control-Allow-Methods", "POST, GET, PUT, DELETE, PATCH")
 	resp.Header().Set("Access-Control-Allow-Credentials", "true")
 
--- a/shared.go
+++ b/shared.go
@@ -17600,6 +17600,8 @@ func handleAgentDecisionStreamResult(workflowExecution WorkflowExecution, action
 		log.Printf("[DEBUG][%s] Got decision ID '%s' for agent '%s'. Ref: %s", workflowExecution.ExecutionId, decisionId, actionResult.Action.ID, actionResult.Status)
 	}
 
+	ctx := context.Background()
+
 	foundActionResultIndex := -1
 	for actionIndex, result := range workflowExecution.Results {
 		if result.Action.ID == actionResult.Action.ID {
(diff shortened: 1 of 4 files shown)


shared.go
```diff
@@ -21415,8 +21415,16 @@ func PrepareSingleAction(ctx context.Context, user User, appId string, body []by
 		workflowExecution.OrgId = user.ActiveOrg.Id
 	}
 
+	formattedAppName := strings.ReplaceAll(strings.ToLower(app.Name), " ", "_")
+
+	isInternalShuffleApp := false
+	switch formattedAppName {
```
frikky: Something about just matching name seems off, but I'm not sure exactly why.

I wonder if there is a way to "steal" the URL field/apikey by formatting an app in a certain way. 

Just need to think it over before merging 🤔

shared.go
```diff
@@ -17101,7 +17101,15 @@ func handleAgentDecisionStreamResult(workflowExecution WorkflowExecution, action
 		// Handle agent decisionmaking. Use the same
 		log.Printf("[INFO][%s] With the agent being finished, we are asking it whether it would like to do anything else", workflowExecution.ExecutionId)
 
-		returnAction, err := HandleAiAgentExecutionStart(workflowExecution, actionResult.Action, true)
+		var originalAction Action
+		if foundActionResultIndex >= 0 && foundActionResultIndex < len(workflowExecution.Results) {
```
frikky: What was the point of this? I'm always confused when we override actions, when things used to work without it

satti-hari-krishna-reddy: At line 8295 in ai.go, we are building actionResult, and inside of that we are setting the action's label name as fmt.Sprintf("Agent Decision %s", decision.RunDetails.Id). Then we call handleAgentDecisionStreamResult, and that one passes actionResult to HandleAiAgentExecutionStart. Now startNode is overwritten with an action with label, e.g., "Agent Decision RNnBU2aY". Due to this, the later nodes in workflow that are trying to refer to the AI agent node with its original label can't find it, resulting in being unable to get any data from the agent node at all.

shared.go
satti-hari-krishna-reddy: Yeah, I think having a special verified field for apps built by support users would be better. Given that we're already overriding the URL field when injecting the auth, if we compare additionally to this special field and both are true, we can trust it. What do you think?

satti-hari-krishna-reddy:
@copilot explain the changes made in this PR

PR #351, opened 2026-03-18
fix: handle raw LLM content with fallback

satti-hari-krishna-reddy:


frikky:
Is this still applicable?

PR #355, opened 2026-03-23
feat: Added RunOpsAgent for agent health check

satti-hari-krishna-reddy:
Added `RunOpsAgent` to the health check function to monitor agent health. Here's how it works:
First, we ask the agent to get the weather info for a city using an API like `wttr`. Then we make our own API call to `wttr` in the background, get the result, and compare it with the agent's output. If the temperature (°C) matches closely with what we got, then the agent worked, so we consider it passing the health check. In all other cases (agent not reachable, failing tool calls, or getting the answer wrong for such a basic task), it's considered a failure.


frikky:
@satti-hari-krishna-reddy @LalitDeore what we doin? :))

LalitDeore:
merging this PR after testing :)

PR #371, opened 2026-04-14
[feature] - Add feature of promoting suborg to parent org

LalitDeore:
When a sub-organization is promoted to a parent organization:

1. Auth, datastore, files, and environment data that were shared from the original parent organization are no longer available.
2. Distributed workflows are detached from the original parent organization. A copy of each distributed workflow is stored in the sub-organization and remains accessible after the promotion.
3. All caches related to workflows, users, organization data, files, auth, and datastore are cleared in both the organization region and the main region.

Note: This feature is available only to support users initially.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -13278,7 +13278,7 @@ func HandlePromoteSubOrg(resp http.ResponseWriter, request *http.Request) {
 		return
 	}
 
-	if requestData.OrgId != "" && requestData.OrgId != parentOrgId {
+	if user.ActiveOrg.Id != parentOrgId {
 		resp.WriteHeader(400)
 		resp.Write([]byte(`{"success": false, "reason": "org_id mismatch between path and body"}`))
 		return
@@ -13369,17 +13369,17 @@ func HandlePromoteSubOrg(resp http.ResponseWriter, request *http.Request) {
 	parentOrg.ChildOrgs = newChildOrgs
 	subOrg.Description = ""
 
-	if err := SetOrg(ctx, *subOrg, subOrg.Id); err != nil {
-		log.Printf("[ERROR] Failed writing promoted org '%s': %s", subOrg.Id, err)
-		resp.WriteHeader(500)
-		resp.Write([]byte(`{"success": false, "reason": "Failed updating sub-organization"}`))
-		return
-	}
-
 	if err := SetOrg(ctx, *parentOrg, parentOrg.Id); err != nil {
 		log.Printf("[ERROR] Failed writing parent org '%s' while promoting '%s': %s", parentOrg.Id, subOrg.Id, err)
 		resp.WriteHeader(500)
 		resp.Write([]byte(`{"success": false, "reason": "Sub-organization promoted, but parent update failed"}`))
+		return
+	}
+
+	if err := SetOrg(ctx, *subOrg, subOrg.Id); err != nil {
+		log.Printf("[ERROR] Failed writing promoted org '%s': %s", subOrg.Id, err)
+		resp.WriteHeader(500)
+		resp.Write([]byte(`{"success": false, "reason": "Failed updating sub-organization"}`))
 		return
 	}
 
--- a/shared.go
+++ b/shared.go
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
 
 }
 
+func resolveParentOrgId(subOrg *Org) string {
+	if subOrg == nil {
+		return ""
+	}
+
+	if subOrg.CreatorOrg != "" {
+		return subOrg.CreatorOrg
+	}
+
+	if len(subOrg.ManagerOrgs) > 0 {
+		return subOrg.ManagerOrgs[0].Id
+	}
+
+	return ""
+}
+
+func clearPromoteSubOrgCaches(ctx context.Context, parentOrgId, subOrgId string, parentOrg *Org, subOrg *Org) {
+	DeleteCache(ctx, fmt.Sprintf("Organizations_%s", parentOrgId))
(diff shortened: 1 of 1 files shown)


shared.go
```diff
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
...
+		return
+	}
+
+	if requestData.SuborgId == "" {
+		resp.WriteHeader(400)
+		resp.Write([]byte(`{"success": false, "reason": "suborg_id is required"}`))
+		return
+	}
+
+	if !user.SupportAccess {
```
frikky, line 13293: I'm primarily worried what would happen here if we give this to users, which is.. kind of the point. They should control this themselves.

shared.go
```diff
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
...
+	}
+
+	requestData := promoteRequest{}
+	if err := json.Unmarshal(body, &requestData); err != nil {
+		resp.WriteHeader(400)
+		resp.Write([]byte(`{"success": false, "reason": "Invalid request body"}`))
+		return
+	}
+
+	if requestData.OrgId != "" && requestData.OrgId != parentOrgId {
```
frikky: I think you should compare it with user.ActiveOrg.Id

That way you ENFORCE the use of Org-Id header when running this request, or to be in the correct org (parent) when it is done.

shared.go
```diff
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
...
+			removedSuborg = true
+			continue
+		}
+
+		newChildOrgs = append(newChildOrgs, child)
+	}
+	parentOrg.ChildOrgs = newChildOrgs
+	subOrg.Description = ""
+
+	if err := SetOrg(ctx, *subOrg, subOrg.Id); err != nil {
```
frikky: When doing this kind of "Setting" of values, try to be consistent.

If you checked Parent Org > Sub Org in validation

Then also keep that order (parent -> child) further down. Makes code way easier to read.

shared.go
```diff
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
...
+		Role:      "admin",
+	}
+
+	workflowItems, err := GetAllWorkflowsByQuery(ctx, subOrgUser, 250, "")
+	if err != nil {
+		log.Printf("[WARNING] Failed loading workflows for promoted org '%s': %s", subOrg.Id, err)
+	}
+
+	for _, workflow := range workflowItems {
+		if workflow.ParentWorkflowId == "" {
```
frikky, line 13398: Wouldn't this be the same as "stealing" a workflow? 

Not sure.

shared.go
```diff
@@ -13165,6 +13165,257 @@ func HandleCreateSubOrg(resp http.ResponseWriter, request *http.Request) {
...
+		DeleteCache(ctx, fmt.Sprintf("%s_workflows", orgUser.Id))
+		DeleteCache(ctx, fmt.Sprintf("user_orgs_%s", orgUser.Id))
+		DeleteCache(ctx, fmt.Sprintf("Users_%s", orgUser.ApiKey))
+		DeleteCache(ctx, fmt.Sprintf("%s", orgUser.ApiKey))
+		DeleteCache(ctx, orgUser.Session)
+		DeleteCache(ctx, fmt.Sprintf("session_%s", orgUser.Session))
+	}
+}
+
+func HandlePromoteSubOrg(resp http.ResponseWriter, request *http.Request) {
```
frikky, line 13232: Please document the API right away on the "API" docs. Otherwise we will forget it.

frikky: (this is why I try to not make new APIs if we don't NEED to) :)

frikky (commented):
I'm sure it works fine for support users only, but I'm worried about opening it up. 

It is a cool feature tho, if it works.

shared.go
LalitDeore: Okay, will write doc for it.

shared.go
LalitDeore: Should we make this available to normal users? I am afraid that if someone promotes a sub-organization to a parent organization without understanding the impact, they could lose access to distributed auth, files, datastore, and environment data due to how things are currently set up and this can impact workflows. I would prefer to keep this limited to support users. Let me know in case you want me to make it public for normal users.

shared.go
LalitDeore: We skip workflows where `workflow.ParentWorkflowId == ""`.

If a workflow has a parent org ID, it means the workflow is distributed, so we reset it to empty but we are making copy of workflow when distributed. That's why it is still accessible after promoting to parent org.

shared.go
LalitDeore: I have added doc for it in below PR. 

https://github.com/Shuffle/shuffle-docs/pull/227

shared.go
frikky: They can just move it back then? References aren't lost..

frikky (commented):
And again, I want to emphasise: The usage of distributed systems is lower than 0.1% of our userbase. Literally one in 1k max. And they are power-users. Power-users need power-tools.  

Publishing the system is the whole point. Otherwhise, why have an API at all? That's just code that won't be used..? 

**Which I tried to make it clear:**
1. MANUALLY DO THIS THIS TIME. Why do we need to waste time building an endpoint that won't be used?

... or

2. Build a proper system that can be used by our users at scale to change parent/child relations at will if they are admins. That way it could actually be useful for management

frikky:
Closing as we discussed ;)

PR #378, opened 2026-04-21
feat: Implemented data reduction for agent response handling 

satti-hari-krishna-reddy:
This pull request introduces a data reduction system for agent responses.The changes implement a new mechanism for agents to specify how much data they need ("data_filter") and which fields are required ("fields_needed"), and ensure that only the necessary data is returned to the agent. This is achieved through both backend logic and updates to the agent prompt instructions.

**Key changes include:**

### Data Reduction Logic

* Added a new `ReduceAgentResponseData` function in `ai.go` that processes raw API responses according to the agent's specified `data_filter` ("count", "list", or "full") and `fields_needed`. The function extracts only the required fields, handles large payloads, and provides fallbacks and warnings when requested fields are missing.

### Prompt and Instruction Updates

* Updated the output format example in the prompt to include `data_filter` and `fields_needed`, reinforcing their required usage for fetch/list/search actions.


satti-hari-krishna-reddy:
@copilot apply changes based on the comments in [this thread](https://github.com/Shuffle/shuffle-shared/pull/378#pullrequestreview-4149807649)

satti-hari-krishna-reddy:
Ahh, Nice review.

cloudSync.go
```diff
@@ -2293,6 +2293,17 @@ func RunAgentDecisionAction(execution WorkflowExecution, agentOutput AgentOutput
 			decision.Tool = appname
 		}
 
+		// Reduce the raw response based on what the agent said it actually needs.
+		// Always route non-empty data filters through the reducer so "full" still
+		// benefits from its passthrough/max-size/error-handling logic.
+		if decision.DataFilter != "" {
```
frikky: This is a problem. What if the next prompt requires details, but previous didn't? 

The point is that data reduction is done on INPUT and not on output. We still need the raw output stored. The data cleanup itself happens when we ask the agent "What should we do now?"

Otherwise we end up in a position of double-grabbing data for now good reason. Plus it makes the user think we are getting the wrong data etc

frikky (commented):
This seems like a fundamental misunderstanding of what we are doing

ai.go
```diff
@@ -7179,6 +7179,208 @@ func sendAITokenLimitAlert(ctx context.Context, execution WorkflowExecution, ful
 	}
 }
 
+const maxRawBytes = 8000
+
+// safeRawFallback enforces the 8000-byte token limit to protect the LLM context window.
+func safeRawFallback(raw []byte, reason string) []byte {
```
frikky: We do NOT want to protect the LLM context window. That is not the point. 

We want things to work. At ALL TIMES. A RAW response is fine. It just should happen so rarely that it doesn't matter.

Why do you keep looking to enshittify what we do instead of looking for continuous ways to make it better?

ai.go
```diff
@@ -7424,9 +7626,19 @@ func HandleAiAgentExecutionStart(execution WorkflowExecution, startNode Action,
...
+				if mappedDecision.DataFilter != "" {
+					originalLen := len(mappedDecision.RunDetails.RawResponse)
+					reduced := ReduceAgentResponseData([]byte(mappedDecision.RunDetails.RawResponse), mappedDecision.DataFilter, mappedDecision.FieldsNeeded)
+					mappedResult.Decisions[i].RunDetails.RawResponse = string(reduced)
+					if debug {
+						log.Printf("[DEBUG][%s] AI_AGENT_REDUCE: decision=%s tool=%s data_filter=%s fields=%v original_bytes=%d reduced_bytes=%d", execution.ExecutionId, mappedDecision.RunDetails.Id, mappedDecision.Tool, mappedDecision.DataFilter, mappedDecision.FieldsNeeded, originalLen, len(reduced))
+					}
+				} else if len(mappedDecision.RunDetails.RawResponse) > 5000 {
+					// No filter set, truncate so the prompt doesn't blow up.
+					mappedResult.Decisions[i].RunDetails.RawResponse = mappedDecision.RunDetails.RawResponse[:5000] + "..."
```
frikky: Random truncating will always be a problem. We don't want to randomly do this.

ai.go
```diff
@@ -7424,9 +7626,19 @@ func HandleAiAgentExecutionStart(execution WorkflowExecution, startNode Action,
...
 				}
 
-				// Truncating, as most valuable details are at the start anyway
-				if len(mappedDecision.RunDetails.RawResponse) > 5000 {
-					mappedDecision.RunDetails.RawResponse = mappedDecision.RunDetails.RawResponse[:5000] + "..."
+				// Reduce data for the LLM prompt
+				if mappedDecision.DataFilter != "" {
+					originalLen := len(mappedDecision.RunDetails.RawResponse)
+					reduced := ReduceAgentResponseData([]byte(mappedDecision.RunDetails.RawResponse), mappedDecision.DataFilter, mappedDecision.FieldsNeeded)
+					mappedResult.Decisions[i].RunDetails.RawResponse = string(reduced)
```
frikky: Doesn't this still fundamentally map update the decision? Or is it just updating the data we send to the agent on this specific request?

What about next actions that may require the full data?

frikky (commented):
Some more things to think about.

I hate that we keep looking for ways to make it all even more shitty. That really is not the point, and I will try to keep making that clear.

Cost reduction is supposed to be done WHILE retaining full functionality. This isn't it.

PR #379, opened 2026-04-22
Fixed the Agent continuation issue

satti-hari-krishna-reddy:


ai.go
```diff
@@ -7462,6 +7462,31 @@ func HandleAiAgentExecutionStart(execution WorkflowExecution, startNode Action,
...
+				if mappedDecision.Action != "ask" {
+					continue
+				}
+				foundContinuation := false
+				for _, field := range mappedDecision.Fields {
+					if field.Key == "continue" && len(field.Answer) > 0 {
+						if debug {
+							log.Printf("[DEBUG][%s] AI Agent continuation: overriding userMessage with 'continue' answer (length=%d)", execution.ExecutionId, len(field.Answer))
+						}
+						userMessage = field.Answer
```
frikky, line 7480: Isn't this a problem?

I'm thinking: how do we deal with original goals vs new goals? If you type something new, should it overwrite the original goal completely? 

Or should it just help direction-change? 

Just curious if that's what we're doing here 🤔

PR #380, opened 2026-04-23
make usage alert cache timeout dymanic

LalitDeore:
**Issues**

1.  In some cases, the same alert is sent multiple times for a single threshold, even though `email_send` is already marked as true in the db.
2.  Alert not being added in the org_stats when app run limit is changed for example from 2K to 10K

**Fix:** 
1. Instead of depending on the database, use a cache and set its timeout to for the remaining days of the month for that alert.
2. When org app runs limit is changed in this case make sure to add alert for changed app runs in org_statistics by making sure both percentage and app runs counts for match in org_stats. Curently, it is checking only any one of them.


frikky:
Neat

PR #395, opened 2026-05-20
Some minor fixes

satti-hari-krishna-reddy:


satti-hari-krishna-reddy:
@frikky Take a look at this when you got some time

PR #401, opened 2026-06-01
fix- multiple issues in backend

LalitDeore:
Fixes:

1. Store the AI usage alert email boolean in the database instead of the cache.
2. Send the last 365 days of data in the Get Stats API instead of only the last 60 days.
3. Do not trim daily statistics to the last 60 days if pushing stats to the GCS bucket fails for any reason. This is safety mechanism to prevent losing past statistics.


0x0elliot (changes requested):
this won't even compile! please remove the duplicate variables

ai.go
```diff
@@ -7187,8 +7189,19 @@ func sendAITokenLimitAlert(ctx context.Context, execution WorkflowExecution, ful
 	aiPercentage := float64(monthlyTokensUsed) / float64(tokenLimit) * 100
 
 	pctCacheKey := generateAlertCacheKey(billingOrgId, fmt.Sprintf("ai_token_pct_%d", int64(100)), admins)
-	if !checkAndSetAlertCache(ctx, pctCacheKey) {
-		log.Printf("[DEBUG] Skipping duplicate AI token alert for org %s, threshold %d%% - already sent recently", billingOrgId, int64(aiPercentage))
+	if !checkAndSetAlertCache(ctx, cacheKey) {
```
0x0elliot, line 7192: cacheKey doesn't even exist

0x0elliot (changes requested):

PR #402, opened 2026-06-02
fix: Make the agent continuation stuff work again

satti-hari-krishna-reddy:


codegen.go
```diff
@@ -4950,7 +4950,16 @@ func handleRunDatastoreAutomation(ctx context.Context, cacheData CacheKeyData, a
 				continue
 			}
 
+			requiredApps := []string{"internal_datastore", "shuffle-datastore"}
```
frikky, line 4953: I made a function for this in blobs.go that should be used:  https://github.com/Shuffle/shuffle-shared/blob/52b913ebf9eb969d84a97e33fad98c76e7bca2d6/blobs.go#L22

frikky (commented):
hi

PR #413, opened 2026-06-19
Several Agent fixes

satti-hari-krishna-reddy:
This PR includes several different fixes for the agent:
1. Fixed an agent parsing bug that was breaking runs. Added multiple fallbacks to improve reliability.
2. Resolved a continuation issue where the LLM would forget the original user request.
3. Made updates to the agent workflow building function.


shared.go
```diff
@@ -38256,20 +38270,94 @@ func broadcastToStream(workflowID string, operation WorkflowOperation, userID st
...
-		Username:  username,
-		Data:      operation.Data,
-		Timestamp: time.Now().UnixMilli(),
+	var streamOps []StreamWorkflowOperation
+
+	for _, operation := range operations {
+		item := "node" // default
+		opType := ""
+		switch operation.Op {
+		case "add_node":
```
0x0elliot: got a feeling that we could have done this more smartly. 

would have been so much more cleaner if case could simply match up with item + opType. Is that possible?

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+// extractDecisionArray finds the first '[' that decodes into a valid decision array.
+func extractDecisionArray(s string) ([]AgentDecision, error) {
+	for i := 0; i < len(s); i++ {
+		if s[i] != '[' {
+			continue
+		}
+
+		// So lets decode into raw maps so we can inspect and mutate it
+		var raw []map[string]interface{}
+		if err := json.NewDecoder(strings.NewReader(s[i:])).Decode(&raw); err != nil {
```
frikky: This kind of stuff is scary. Always split into multiple at a time if possible, and verify indexing.

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+			continue
+		}
+
+		if len(raw) == 0 || raw[0]["action"] == nil {
```
frikky: What happens if "action" doesn't exist in the first object?

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+			continue
+		}
+
+		// Fix the types (convert numbers/bools to strings)
+		normalizeDecisionFields(raw)
```
frikky: If it's gonna normalize, doesn't it need to return &raw? 

Or are pointer mappings enough?

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+func extractDecisionJSONL(s string) ([]AgentDecision, error) {
+	var out []AgentDecision
+
+	for i := 0; i < len(s); {
+		if s[i] != '{' {
+			i++
+			continue
+		}
+
+		r := strings.NewReader(s[i:])
```
frikky: Why `s[i:]`? Just to read from current index until end? Feels weird without a moving window

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+		dec := json.NewDecoder(r)
+
+		var raw map[string]interface{}
+		if err := dec.Decode(&raw); err != nil {
+			i++
+			continue
+		}
+
+		if raw["action"] != nil {
```
frikky: Yay, validation

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+			normalizeDecisionFields([]map[string]interface{}{raw})
```
frikky: Try to avoid this kind of pointer magic please. And pre-declare values. 

This isn't easily understandable/readable

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+// This MUST happen before unmarshalling into AgentDecision (whose Value field is a string).
+func normalizeDecisionFields(decisions []map[string]interface{}) {
+	for _, d := range decisions {
+		fields, ok := d["fields"].([]interface{})
+		if !ok {
+			continue
+		}
+
+		for _, f := range fields {
+			fm, ok := f.(map[string]interface{})
```
frikky: I'm scared when we do casting -> check values (:

ai.go
```diff
@@ -1706,6 +1673,136 @@ func balanceJSONLikeString(s string) string {
...
+	}
+
+	return out, nil
+}
+
+// normalizeDecisionFields converts non-string field["value"] entries into JSON strings.
+// This MUST happen before unmarshalling into AgentDecision (whose Value field is a string).
+func normalizeDecisionFields(decisions []map[string]interface{}) {
+	for _, d := range decisions {
+		fields, ok := d["fields"].([]interface{})
```
frikky: Don't cast AND check if it exists at the same time. Separate.

frikky (commented):
I'm scared of how many potential nil-pointers you may have introduced.

4 lines doing 1 thing each >>>>> 1 line doing 4 things

PR #416, opened 2026-06-23
New endpoint for the workflow editor

Monilprajapati:


shared.go
```diff
@@ -37473,28 +37473,12 @@ func ListProcesses() ([]ProcessInfo, error) {
...
-	ctx := GetContext(request)
-	user, userErr := HandleApiAuthentication(resp, request)
-	if userErr != nil {
-		log.Printf("[WARNING] Api authentication failed in GetWorkflowAppsSummary: %s", userErr)
-		resp.WriteHeader(401)
-		resp.Write([]byte(`{"success": false}`))
-		return
-	}
-
+func getOrgAppSummaries(ctx context.Context, user User) ([]AppSummary, error) {
```
satti-hari-krishna-reddy, line 37476: This function is not returning the summaries but rather the entire blob of the app information ?

Monilprajapati: No, it's only returning the app summaries. Check the return struct (Name, Description and Id)

PR #421, opened 2026-06-29
fix: refresh OpenSearch bulk writes for immediate read-after-write

0x0elliot:


0x0elliot:
fixes the minor issue we had to deliver fast. take a quick look once

PR #422, opened 2026-06-29
fix: refresh OpenSearch deletes for immediate read-after-delete

0x0elliot:


0x0elliot:
it's just a one line change, so i am merging it for now. i need to get back to someone with it.

PR #426, opened 2026-07-07
Feat/vulnerability correlation

0x0elliot:


blobs.go
```diff
@@ -1186,6 +1186,206 @@ func GetDefaultWorkflowByType(workflow Workflow, orgId string, categoryAction Ca
...
+					AppName:     "Shuffle Tools",
+					ID:          startActionId,
+					AppVersion:  "1.2.0",
+					Environment: actionEnv,
+					Label:       "verify_package_vulnerability",
+					IsStartNode: true,
+					Sharing:     true,
+					Parameters: []WorkflowAppActionParameter{
+						{
+							Name:      "code",
```
frikky: Would be great if this was not inline to keep this area readable. Just make a function for it like we do elsewhere.

0x0elliot: cool. added.

blobs.go
```diff
@@ -567,6 +567,204 @@ func HandleSingulWorkflowEnablement(ctx context.Context, workflow Workflow, user
 // That means each algorithm needs to be written as if-statements to
 // replace a specific part of a workflow :thinking:
 
+func getVulnerabilityCorrelationWorkflow(actionType, orgId, startActionId, actionEnv string, categoryAction CategoryAction) Workflow {
```
frikky, line 570: The whole point is that putting a single Python script within a bunch of Golang code is absolutely horrendous. 

Do not write code without looking at how things are already please. You are not the first to do this, so we have a structure already. If you disagree, at least tell me, so we can find the correct solution and make it identical everywhere. This is just sloppy.

frikky (commented):
come on...

frikky:
Merging and fixing on my end

PR #443, opened 2026-07-31
feat: Added support for running direct tool calls by the agent

satti-hari-krishna-reddy:
## What does this PR do?

- Makes it optional for tool calls to use Singul.
- Calls executeSingleAction directly, which calls PrepareSingleAction to run the actual app.
- This avoids the overhead of using Singul in cases where the agent already knows the app action's schema and passes the correct data.
- The Singul bypass is controlled by injecting an execution_mode=direct parameter from the agent caller function.

### DIFF
--- a/cloudSync.go
+++ b/cloudSync.go
@@ -1987,12 +1987,14 @@ func HandleSuborgScheduleRun(request *http.Request, workflow *Workflow) {
 // runAgentDecisionDirectAppCall bypasses Singul and runs the app directly.
 func runAgentDecisionDirectAppCall(execution WorkflowExecution, decision AgentDecision) (rawResult []byte, debugUrl string, appName string, categoryLabels []string, actionName string, err error) {
 	ctx := context.Background()
-	minUser := User{
-		Username: execution.Workflow.Owner,
-		ActiveOrg: OrgMini{
-			Id: execution.ExecutionOrg,
-		},
+	var minUser User
+	org, err := GetOrg(ctx, execution.ExecutionOrg)
+	if err == nil && len(org.Users) > 0 {
+		minUser = org.Users[0] // Grabbing the first user, Is this the right way to do it ?
+	} else {
+		minUser = User{Username: execution.Workflow.Owner}
 	}
+	minUser.ActiveOrg = OrgMini{Id: execution.ExecutionOrg}
 
 	var (
 		resolvedAppId   string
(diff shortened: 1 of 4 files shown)


cloudSync.go
```diff
@@ -1984,21 +1984,242 @@ func HandleSuborgScheduleRun(request *http.Request, workflow *Workflow) {
 	}
 }
 
+// runAgentDecisionDirectAppCall bypasses Singul and runs the app directly.
+func runAgentDecisionDirectAppCall(execution WorkflowExecution, decision AgentDecision) (rawResult []byte, debugUrl string, appName string, categoryLabels []string, actionName string, err error) {
+	ctx := context.Background()
+	minUser := User{
+		Username: execution.Workflow.Owner,
```
frikky: This won't necessarily work long-term. Using org -> lookup user somehow is almost always better

cloudSync.go
```diff
@@ -1984,21 +1984,242 @@ func HandleSuborgScheduleRun(request *http.Request, workflow *Workflow) {
...
+		} else {
+			port := os.Getenv("PORT")
+			if len(port) == 0 {
+				port = "5001"
+			}
+			baseURL = fmt.Sprintf("http://localhost:%s", port)
+		}
+	}
+
+	requestUrl := fmt.Sprintf("%s/api/v1/apps/%s/run?delete=false&execution_id=%s&authorization=%s&org_id=%s&timeout=115", baseURL, resolvedAppId, execution.ExecutionId, execution.Authorization, execution.ExecutionOrg)
```
frikky: Why such a high timeout? Is it necessary? And why so specific at 115?

Most of our apps time out at 60 seconds, or even 30 seconds by default. That is the app itself.

cloudSync.go
```diff
@@ -1984,21 +1984,242 @@ func HandleSuborgScheduleRun(request *http.Request, workflow *Workflow) {
...
 	// Check if running in test mode
 	if os.Getenv("AGENT_TEST_MODE") == "true" {
 		log.Printf("[DEBUG][%s] AGENT_TEST_MODE enabled - using mock tool execution", execution.ExecutionId)
-
-		// Call mock handler
 		body, debugUrl, appName, err := RunAgentDecisionMockHandler(execution, decision)
 		return body, debugUrl, appName, []string{}, "", err
 	}
 
+	skipSingul := executionMode == "direct" || os.Getenv("AGENT_SKIP_SINGUL") == "true"
```
frikky, line 2217: ah, neat. 

We need to clean up environment variables soon lol

frikky (commented):
Just some small-ish things.

PR #457, opened 2026-08-17
Add error handling for JSON marshaling in buildWorkflowEditContext

satti-hari-krishna-reddy:


Monilprajapati:
Found that @frikky has already made the change for it :)

PR #458, opened 2026-08-19
fix:  Prevent agent infinite loops on cache failure with DB fallback guard

satti-hari-krishna-reddy:


shared.go
```diff
@@ -17787,8 +17787,23 @@ func sendAgentActionSelfRequest(status string, workflowExecution WorkflowExecuti
 
 		cacheErr := SetCache(ctx, cacheKey, []byte("1"), cacheTTL)
 		if cacheErr != nil && (status == "SUCCESS" || status == "FINISHED" || status == "FAILURE" || status == "ABORTED") {
-			log.Printf("[WARNING][%s] Memcache down — skipping agent self-request for '%s' to prevent retry storm", workflowExecution.ExecutionId, status)
-			return nil
+			log.Printf("[WARNING][%s] Cache error, falling back to DB check for agent self-request '%s'", workflowExecution.ExecutionId, status)
+			
+			// DB Fallback
+			dbExec, err := GetWorkflowExecution(ctx, workflowExecution.ExecutionId)
```
frikky: Does DB fallback actually work here, as it also has a caching element?

frikky: Question is: Should we add an optional "SkipCache" parameter to the GetWorkflowExecution function?

satti-hari-krishna-reddy: Yes, we should add it to guarantee it reads directly from the database.

PR #461, opened 2026-08-26
Fix sso login issue for new users when sso is required for org

LalitDeore:


shared.go
```diff
@@ -16336,7 +16336,15 @@ func HandleLogin(resp http.ResponseWriter, request *http.Request) {
 				log.Printf("[INFO] OpenID login for %s", org.Id)
 				redirectKey = "SSO_REDIRECT"
 
-				baseSSOUrl, err = GetOpenIdUrl(request, *org, userdata, "signin")
+				mode := "signin"
```
0x0elliot, line 16339: i get the point of doing this stuff with the login API but for the users you're making this fix for, this won't work out :)

0x0elliot (changes requested):

0x0elliot:
looks good to me!

PR #462, opened 2026-08-28
A lot of agent fixes

satti-hari-krishna-reddy:


shared.go
```diff
@@ -83,6 +83,8 @@ import (
 )
 
 var project ShuffleStorage
+var agentNextLLMCallMu sync.Map // Mutex to prevent duplicate parallel agent LLM runs
```
frikky: Sooo, how does this work, not just in multithreaded systems, but multi-server systems?

frikky (commented):
(:

shared.go
satti-hari-krishna-reddy: Well, it won't work on a multi server system since it's stored on ram. I replaced it with cache so it works for the same purpose.

shared.go
```diff
@@ -18311,7 +18316,15 @@ func handleAgentDecisionStreamResult(workflowExecution WorkflowExecution, action
 			originalAction = actionResult.Action
 		}
 
-		// If the execution is already FINISHED but a continuation was injected (user asked  the agent to do more on top of what it already did), we need to reset the status back to EXECUTING so that HandleAiAgentExecutionStart doesn't exit with "Agent run already finished". We also persist this to cache so the guard in
+		lockKey := fmt.Sprintf("agent_llm_lock_%s_%s", workflowExecution.ExecutionId, originalAction.ID)
```
frikky, line 18319: I got a feeling the lock mechanism will cause issues. 

Do we have a rerun mechanism in case we get stuck in a thread?

frikky (commented):
Merging

PR #464, opened 2026-09-03
Limits & Billing Changes

LalitDeore:


ai.go
```diff
@@ -9906,31 +9893,23 @@ data_filter:
...
+						currentOrgId = billingOrgId
+					}
 
-					subOrgId := execution.Workflow.OrgId
 					go func() {
 						time.Sleep(time.Duration(rand.Intn(500)) * time.Millisecond)
+						IncrementCache(ctx, currentOrgId, "agent_tokens", totalTokens)
 						if inputTokens > 0 {
-							IncrementCache(ctx, billingOrgId, "agent_input_tokens", inputTokens)
+							IncrementCache(ctx, currentOrgId, "agent_input_tokens", inputTokens)
```
frikky: wtf is this? 

Why are we counting agent tokens here?

blobs.go
```diff
@@ -2597,6 +2597,14 @@ func GetBrandingAvailable(key string) bool {
 	return false
 }
 
+func GetAppRunsGrouping(key string) bool {
+	if key == "1e1bf9b426033f9f15e8070e007f9414d11d7fdd6a07402a9290e1a0d7965f8f" {
```
frikky, line 3202: What is this?

db-connector.go
```diff
@@ -8171,15 +8323,78 @@ func GetWorkflowQueue(ctx context.Context, id string, limit int, inputEnv ...Env
 
 		license := checkNoInternet()
 		if license.Valid {
-			limit = limit * 2
+			if license.AppRunsGrouping {
```
frikky, line 8318: What is this grouping thing? Is it another "counter" that is separate?

cloudSync.go
```diff
@@ -918,30 +918,96 @@ func ValidateExecutionUsage(ctx context.Context, orgId string) (*Org, error) {
...
 		if err != nil {
-			log.Printf("[ERROR] Failed setting org statistics for monthly reset for %s (%s): %s ", validationOrg.Name, validationOrg.Id, err)
+			log.Printf("[ERROR] Failed setting org statistics after daily rollover for %s (%s): %s ", validationOrg.Name, validationOrg.Id, err)
 		}
 	}
 
 	totalAppExecutions := validationOrgStats.MonthlyAppExecutions + validationOrgStats.MonthlyChildAppExecutions
-	if validationOrg.Billing.InternalAppRunsHardLimit > 0 && totalAppExecutions > validationOrg.Billing.InternalAppRunsHardLimit {
-		return validationOrg, errors.New(fmt.Sprintf("Org %s (%s) has exceeded app runs hard limit (%d/%d) - Only Shuffle Support can control this metric.", validationOrg.Name, validationOrg.Id, totalAppExecutions, validationOrg.Billing.InternalAppRunsHardLimit))
+	if validationOrg.SyncFeatures.AnnualAppRunsGrouping.Active == false && validationOrg.Billing.InternalAppRunsHardLimit > 0 && totalAppExecutions > validationOrg.Billing.InternalAppRunsHardLimit {
```
frikky, line 931: I'm so confused about this app grouping thing

db-connector.go
```diff
@@ -916,6 +913,86 @@ func IncrementCacheDump(ctx context.Context, orgId, dataType string, amount ...i
 		}
 	}
 
+	if len(tmpOrgDetail.ManagerOrgs) > 0 && (dataType == "agent_tokens") {
```
frikky, line 916: Didn't we talk about LLM tokens being the root? 🤦 

Agent tokens just describe agents. LLMs describe EVERYTHING. That means we should use LLM as the base.

db-connector.go
```diff
@@ -3841,6 +3918,60 @@ func GetOrgByCreatorId(ctx context.Context, id string) (*Org, error) {
...
+				Percentage: percentage,
+				Count:      int(float64(percentage) / 100 * float64(limit)),
+			})
+		}
+	}
+
+	return thresholds
+}
+
+func addDefaultAlertThresholds(org *Org) bool {
```
frikky, line 3959: Why apply them? Why not dynamically just use them instead? And ignore the if the user sets their own limits?

The problem with actually setting them is just that: They are then set. Meaning any future change require database updates, and not just a backend mod

db-connector.go
```diff
@@ -8136,11 +8283,16 @@ func GetWorkflowQueue(ctx context.Context, id string, limit int, inputEnv ...Env
 
 	if project.Environment != "cloud" && len(inputEnv) > 0 && len(executions) > 0 {
 		env := inputEnv[0]
-
 		orgId := env.OrgId
-		org, err := GetOrg(ctx, orgId)
+
+		org, err := GetFirstOrg(ctx)
```
frikky, line 8280: I'm confused. Why is this here? Why are we still looking for an orgId? 

This seems like an exploit waiting to happen

frikky (commented):
Added a few comments. Check em please, and get back to me

ai.go
LalitDeore: Should we count both LLM and agent token as well then?

blobs.go
LalitDeore: This is for the annual app run grouping. For air gap users when this key exist in their license key then we combine all of the annual app run limi to single limit and limitation happens base on it's annual app runs instead of monthly app runs.

cloudSync.go
LalitDeore: For onprem and cloud users when this feature is enabled then we combine all of the annual app run limit to single limit and limitation happens base on it's annual app runs instead of monthly app runs.

stats.go
```diff
@@ -655,6 +656,50 @@ func GetSpecificStats(resp http.ResponseWriter, request *http.Request) {
 	resp.Write([]byte(fmt.Sprintf(`{"success": %v, "key": "%s", "total": %d, "available_keys": %s, "entries": %s}`, successful, strings.ReplaceAll(statsKey, "\"", ""), totalValue, string(availableStats), string(marshalledEntries))))
 }
 
+func mergeMultiRegionResults(crossRegionResults []MultiRegionStatsEntry, info *ExecutionInfo, parentDailyMap map[string]int) {
```
frikky, line 659: Is this persistent?

stats.go
```diff
@@ -655,6 +656,50 @@ func GetSpecificStats(resp http.ResponseWriter, request *http.Request) {
...
+			alreadyAppliedCorrection := childDay.AgentInputTokens*250/1_000_000 +
+				childDay.AgentOutputTokens*1500/1_000_000 +
+				childDay.DailySMSUsage*3 +
+				childDay.DailyEmailUsage*2
+			rawChildAppExecutions := childDay.AppExecutions - alreadyAppliedCorrection
+			if rawChildAppExecutions < 0 {
+				rawChildAppExecutions = 0
+			}
+
+			if idx, exists := parentDailyMap[dateKey]; exists {
```
frikky, line 673: This feels funky. You aren't checking if info.DailyStaticstics contains idx..? Will lead to nil pointer.. somehow

db-connector.go
LalitDeore: Yes yes, this is old code in pending from last 3 month. I am working on changing it to the LLM token. But should we count both LLM token and agents token while counting agent token? So, this there will be separation of the LLM and agent tokens

stats.go
```diff
@@ -835,6 +880,54 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+		parsedItem := []byte(cacheItem.([]uint8))
+		increment, err := strconv.Atoi(string(parsedItem))
+		if err == nil {
+			info.TotalChildOrgEmailUsage += int64(increment)
+			info.MonthlyChildOrgEmailUsage += int64(increment)
+			info.DailyChildOrgEmailUsage += int64(increment)
+		}
+	}
+
+	key = fmt.Sprintf("cache_%s_send_sms", orgId)
```
frikky, line 906: What is this SMS stuff for?

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
 		}
 	}
 
-	newjson, err := json.Marshal(info)
+	parentOrgLocations := []Locations{}
+	parentEnvs, err := GetEnvironments(ctx, org.Id)
```
frikky, line 966: What is this for..?

db-connector.go
LalitDeore: This license.AppRunsGrouping is for the air gap users which is true when license key for annual app runs grouping is true.

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+		for i, d := range info.DailyStatistics {
+			parentDailyMap[d.Date.UTC().Format("2006-01-02")] = i
+		}
+
+		// mu protects all shared writes: info.Tenants, info.Locations, info.DailyStatistics, parentDailyMap
+		var mu sync.Mutex
+		var wg sync.WaitGroup
+
+		// Semaphore: buffered channel of size 5 limits concurrent goroutines to 5 at a time
+		sem := make(chan struct{}, 5)
```
frikky, line 1012: Does it actually do 5 at a time? Like if I have 500, it still does 5 at a time?

If I have 2 child-tenants, does it keep the rest of those buffered channels open anyway, as in unassigned memory? That's how you get crashes and takeovers (:

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+				mu.Lock()
+				info.Tenants = append(info.Tenants, newTenant)
+				info.Locations = append(info.Locations, newLocs...)
+				mu.Unlock()
+			}(childOrgMini)
+		}
+
+		wg.Wait()
+
+		if project.Environment == "cloud" && hasCrossRegionChildren && !skipMultiRegion {
```
frikky, line 1086: I guess this crossRegionChildren is my largest worry in this whole release. Exactly how we're dealing with the data from them, that is.

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+					}
+				}
+
+				childRegionUrl := strings.TrimRight(childOrgFull.RegionUrl, "/")
+				if project.Environment == "cloud" &&
+					len(childRegionUrl) > 0 &&
+					strings.Contains(childRegionUrl, "http") &&
+					childRegionUrl != parentRegionUrl {
+					mu.Lock()
+					hasCrossRegionChildren = true
```
frikky, line 1070: Does this always set the variable in the main thread even when this is ran in a goroutine?

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+
+		if project.Environment == "cloud" && hasCrossRegionChildren && !skipMultiRegion {
+			multiRegionCacheKey := fmt.Sprintf("multi_region_stats_%s", org.Id)
+
+			if cachedBody, cacheErr := GetCache(ctx, multiRegionCacheKey); cacheErr == nil {
+				cachedBytes := []byte(cachedBody.([]uint8))
+				var cachedResults []MultiRegionStatsEntry
+				if jsonErr := json.Unmarshal(cachedBytes, &cachedResults); jsonErr == nil {
+					log.Printf("[INFO] HandleGetStatistics: serving multi-region stats from cache for org %s", orgId)
+					mergeMultiRegionResults(cachedResults, info, parentDailyMap)
```
frikky, line 1094: I'm worried about this. Was it tested on 500+ child-tenants?

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+				} else {
+					log.Printf("[WARNING] HandleGetStatistics: failed unmarshalling cached multi-region-stats, will refetch: %s", jsonErr)
+				}
+			} else {
+				multiRegionUrl := fmt.Sprintf("https://shuffler.io/api/v1/orgs/%s/multi-region-stats", orgId)
```
frikky: Is this some kind of forwarder then? Like the 3rd party region itself is sending the stats here? How has this been tested then?

PS: Use uk.shuffler.io when it's static, as shuffler.io is a route to uk.shuffler.io

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+						} else {
+							log.Printf("[WARNING] HandleGetStatistics: multi-region-stats returned status %d", multiResp.StatusCode)
+						}
+					}
+				}
+			}
+		}
+	}
+
+	if org.SyncFeatures.AnnualAppRunsGrouping.Active {
```
frikky, line 1136: Why do I keep seeing "Annual" on this? Is it due to license mapping typically being annual?

stats.go
```diff
@@ -870,7 +963,207 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+				endDate = time.Unix(subscription.Enddate, 0)
+				break
+			}
+		}
+
+		if annualSubscriptionExists && len(info.DailyStatistics) > 0 {
+
+			annualSubscriptionAppRuns := int64(0)
+			annualSubscriptionChildAppRuns := int64(0)
+			for i := range info.DailyStatistics {
```
frikky, line 1152: If you are just validating, and not updating anything in the original array, why use indexing like this?

Just use for _, stat := range info.DailyStatistics{}

Way simpler to read than constant indexing.

stats.go
```diff
@@ -882,6 +1175,56 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
 	resp.Write(newjson)
 }
 
+// Make sure that we are not calling SetOrgStatistics function after calling this function. This will increase the app runs count in db on every call to this function.
```
frikky, line 1177: Why is this a worry? Can we prevent that..?

Why would calling SetOrgStatistics create an app run increase..? That sounds like a proper algorithm flaw

stats.go
```diff
@@ -882,6 +1175,56 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+func GetCorrectedStats(info *ExecutionInfo) *ExecutionInfo {
+
+	// 1 Million Input Tokens = 250 app runs
+	// 1 Million Output Tokens = 1500 app runs
+	// 1 SMS = 3 app runs
+	// 1 Email = 2 app runs
+
+	// Loop through the daily statistics and add the app runs from tokens, SMS and email on top of existing counts
+	for i := range info.DailyStatistics {
```
frikky, line 1185: Wait, is this being stored?

Or is it just a calculation?

stats.go
```diff
@@ -882,6 +1175,56 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+	if len(info.DailyStatistics) > 0 {
+		annualInputTokens := int64(0)
+		annualChildInputTokens := int64(0)
+		annualOutputTokens := int64(0)
+		annualChildOutputTokens := int64(0)
+		annualSMSUsage := int64(0)
+		annualChildSMSUsage := int64(0)
+		annualEmailUsage := int64(0)
+		annualChildEmailUsage := int64(0)
+		for i := range info.DailyStatistics {
```
frikky, line 1207: Don't use indexing when you don't update the current array.. (:

Readability please

stats.go
```diff
@@ -882,6 +1175,56 @@ func HandleGetStatistics(resp http.ResponseWriter, request *http.Request) {
...
+func GetCorrectedStats(info *ExecutionInfo) *ExecutionInfo {
+
+	// 1 Million Input Tokens = 250 app runs
+	// 1 Million Output Tokens = 1500 app runs
+	// 1 SMS = 3 app runs
+	// 1 Email = 2 app runs
+
+	// Loop through the daily statistics and add the app runs from tokens, SMS and email on top of existing counts
+	for i := range info.DailyStatistics {
+		info.DailyStatistics[i].AppExecutions += info.DailyStatistics[i].AgentInputTokens*250/1_000_000 + info.DailyStatistics[i].AgentOutputTokens*1500/1_000_000 + info.DailyStatistics[i].DailySMSUsage*3 + info.DailyStatistics[i].DailyEmailUsage*2
```
frikky: PS: Here indexing makes sense.

frikky (commented):
More.. things to consider

stats.go
LalitDeore: No, stats merging happens in memory only

stats.go
LalitDeore: Didn't get it. It shouldn't cause any nil pointer.  We are already using the info.DailyStatistics[idx] below

stats.go
LalitDeore: We are storing SMS and Email usage in stats api from now. So, it's sms cache

stats.go
LalitDeore: It is for the location stats. We will show all locations & tenants count usage on billing page.

stats.go
LalitDeore: Yes, it does 5 at time even if user have 500 tenants in multiple regions. 

If user have 2 child tenant then in this case it only send 2 go routines and the other 3 buffer slots are just never written to.

stats.go
LalitDeore: No, this variable assignment is inside the condition: 

childRegionUrl != parentRegionUrl

stats.go
LalitDeore: Nope, I have tested it on 6-7 tenants

stats.go
LalitDeore: Okay, changing it to uk.shuffler.io

stats.go
LalitDeore: It is for combining annual app runs so they can use it across the year instead of monthly app runs limit

stats.go
LalitDeore: If we call GetCorrectedStats inside GetOrgStatistics, or call setOrgStatistics after GetCorrectedStats, it will return the original app runs plus the converted app runs from AI token, SMS, and email usage.

If we then call setStatistics, these converted stats will be stored in the DB. When we call GetOrgStatistics again, it will add the converted app runs again, resulting in:

app runs + converted app runs + app runs + converted app runs

stats.go
LalitDeore: No, it is not being stored.

ai.go
frikky: The problem is it's double counting. LLM and Agent tokens are counted in RunAiQuery()

blobs.go
frikky: This key exactly, or the key that the hashing translates it into? 

I still, from an algorithmic perspective, don't understand

db-connector.go
frikky: This is so confusing. Why not just check all 12 months?

db-connector.go
frikky: Yes, that is what the conversation has been since we first talked about it. The separation is to understand where LLM tokens are going.

stats.go
frikky: "Shouldn't" is not the same as "not possible"

To me, this reads as bad validation, as you are validating from parentDailyMap, and taking some random value from it into user.DailyStatistics[idx]

This shouts nil-pointer at me, as there is no explicit check

stats.go
frikky: Another security flaw then. Any non-used memory, allows for potential memory manipulation. You never create larger memory space than what you intend to use. Creating smaller is fine.

That means you only make it 5, if there is a minimum of 5 child-tenants, 4 if 4, 3 if 3 etc.

stats.go
frikky: I'm asking if it's setting it in the main thread, even though the update is happening in a goroutine. And if that works 100% of the time

stats.go
frikky: Gotta expand that testing ;)

ai.go
LalitDeore: Ahh, saw that we are incrementing agent and llm tokens here. But we are only calculating total tokens. I think I should add input, output and cache token for both LLM and agent? 

https://github.com/Shuffle/shuffle-shared/blob/2ad81d2533a6d6d0925e1ddde091610ce07f78bd/ai.go#L11443

blobs.go
LalitDeore: This is hash of the key. I have sent app runs grouping key to business team to store it.

db-connector.go
LalitDeore: I observed that env.OrgId is not actual org id to search it can be like "shuffle" so use GetFirstOrg and then base on that find the actual limits

db-connector.go
LalitDeore: Because, annual app runs will be base on the contract of customers so set key and limits base on it.

stats.go
LalitDeore: I will look into it if this is possible somehow

stats.go
LalitDeore: I will look into it if it can crash somehow and will get back here

stats.go
LalitDeore: No, we are not setting it in the main thread. We are setting it per tenant. hasCrossRegionChildren = true is set inside the go routine but write is inside the mu.LOCK()/Unlock()

PR #467, opened 2026-09-04
[feature] - AI agent execution through the cloud sync

LalitDeore:


ai.go
```diff
@@ -8183,6 +8183,22 @@ func HandleAiAgentExecutionStart(execution WorkflowExecution, startNode Action,
 	}
 
 	// Validate On-Prem Configuration immediately
+	if project.Environment == "onprem" {
+		if os.Getenv("AI_MODEL") == "" && os.Getenv("OPENAI_MODEL") == "" {
```
frikky, line 8187: What is the point of this? We use App Authentication for it - not environment variables.

ai.go
```diff
@@ -15542,12 +15554,23 @@ func GetOrgAiCredentials(ctx context.Context, callInfo AiCallInfo) (string, stri
 		}
 	}
 
-	// Handles failover IF we can't find other auth
-	if project.Environment != "cloud" && (apiKey == "" || aiRequestUrl == "") {
-		log.Printf("[INFO] No custom LLM-credentials found for org %s. Falling back to default shuffler.io AI endpoint IF cloud sync is enabled.", orgId)
+	// Handles failover IF we can't find other auth.
+	// Only applies to on-prem deployments - cloud never needs to fail over to itself.
+	if project.Environment == "onprem" && (apiKey == "" || aiRequestUrl == "") {
+		if apiKey == "" {
```
frikky, line 15560: Same again

ai.go
```diff
@@ -15561,9 +15584,24 @@ func GetOrgAiCredentials(ctx context.Context, callInfo AiCallInfo) (string, stri
 			return "", "", ""
 		}
 
-		// In case it's stored from cloudsync
-		if strings.Contains(org.SyncConfig.URL, "shuffler.io") && strings.HasPrefix(org.SyncConfig.URL, "https://") {
-			baseUrl = org.SyncConfig.URL
+		baseUrl := "https://shuffler.io"
```
frikky: Default to uk.shuffler.io please.

ai.go
LalitDeore: These environments are already present in the cloud. Aren't we using both API key and app authentication? By default, we use the API key or app authentication if either is configured. If Cloud Sync is enabled, then we use Cloud Sync instead and api key  and app auth are missing

ai.go
LalitDeore: Okay

ai.go
frikky: Nope, we moved away from all of this like 3 weeks ago. That was the whole Google Gemini update with multi-region setups. It partially still exists, but will be removed little by little, as UI modifications is WAY better. As it should be.

LalitDeore: Okay, I will remove this environment variable then

PR #470, opened 2026-09-07
Improve workflow trigger handling 

satti-hari-krishna-reddy:


satti-hari-krishna-reddy:
@frikky Please take a look at this

shared.go
```diff
@@ -37587,9 +37616,17 @@ func HandleAgentWorkflowOperations(resp http.ResponseWriter, request *http.Reque
 		// Track triggers being added/deleted so we can start/stop them AFTER
 		// the full workflow is constructed (start node + branches all resolved).
 		var triggerToStop *Trigger
-		if (operation.Op == "delete_node" || operation.Op == "remove_node") && operation.NodeType == "trigger" {
+		if operation.Op == "delete_node" || operation.Op == "remove_node" {
+			targetID := operation.ID
+			if realID, ok := tempIDMap[targetID]; ok {
+				targetID = realID
+			} else if len(targetID) == 0 && len(operation.TempID) > 0 {
```
frikky: If you check for len(targetID) == 0 AFTER checking indexing, what happens if tempIDMap[""] happens? Does that work? Does it crash? It feels like the order here may be wrong

shared.go
```diff
@@ -38214,6 +38304,14 @@ func opAddNodeWithMapping(ctx context.Context, user User, wf *Workflow, op *Work
 		}
 	}
 
+	// Resolve temp_ids in insert_before and insert_after if provided
+	if realID, exists := tempIDMap[op.InsertBefore]; exists {
```
yashsinghcodes: nil pointer here?

shared.go
```diff
@@ -38214,6 +38304,14 @@ func opAddNodeWithMapping(ctx context.Context, user User, wf *Workflow, op *Work
...
+		op.InsertBefore = realID
+	}
+	if realID, exists := tempIDMap[op.InsertAfter]; exists {
```
yashsinghcodes: and here?

PR #476, opened 2026-09-10
support AI Agent execution in on-prem and hybrid environments and 

satti-hari-krishna-reddy:


cloudSync.go
```diff
@@ -2504,24 +2516,29 @@ func runAgentDecisionDirectAppCall(execution WorkflowExecution, decision AgentDe
 		}
 	}
 
-	//ExecutionDelay: selectedDelay,
-	timeout := time.Duration(30) * time.Second
+	toolTimeout := 120
+	if v := os.Getenv("AGENT_TOOL_TIMEOUT"); len(v) > 0 {
```
frikky: This will 100% cause issues.

cloudSync.go
```diff
@@ -3073,7 +3090,10 @@ func RunAgentDecisionAction(execution WorkflowExecution, agentOutput AgentOutput
 	//     2. Send the result through AI again to check if it changes (?). Should there be a verdict here?
 	//     3: Start the next steps of decisions after updates
 
-	baseUrl := "https://shuffler.io"
+	baseUrl := "http://localhost:5001"
+	if project.Environment == "cloud" {
+		baseUrl = "https://shuffler.io"
```
frikky: Uhm, do we pass in CLOUDRUN URL as well? shuffler.io is only used for the main region, and the actual correct one is uk.shuffler.io

And for localhost:5001. That's the default port only, passed by PORT env.

shared.go
```diff
@@ -19213,6 +19234,53 @@ func ParsedExecutionResult(ctx context.Context, workflowExecution WorkflowExecut
 		}
 	}
 
+	// Specific handler for AI Agent hybrid thing (where worker ran run_agent and it reached to Cloud)
+	isAgentAction := actionResult.Action.AppName == "shuffle-ai" || actionResult.Action.AppName == "AI Agent" || actionResult.Action.AppName == "Shuffle Agent" || actionResult.Action.Name == "run_agent"
+	isAgentHybrid := isAgentAction && (strings.Contains(strings.ToLower(actionResult.Result), "hybrid") || actionResult.Action.Name == "run_agent")
+
+	log.Printf("[HEYOOO-PARSED] ParsedExecutionResult: Label='%s', AppName='%s', ActionName='%s', isAgentHybrid=%t, ResultPreview=%.150s", actionResult.Action.Label, actionResult.Action.AppName, actionResult.Action.Name, isAgentHybrid, actionResult.Result)
+
+	if isAgentHybrid && actionResult.Status != "SKIPPED" {
```
frikky: I'm not sure I understand "isAgentHybrid"..?

What is the point of this? The agent itself wouldn't be hybrid? That just runs wherever the backend is. The actions it takes however, can be hybrid

shared.go
```diff
@@ -36796,14 +36880,45 @@ func getPrioritisedAppActions(ctx context.Context, inputApp string, maxAmount in
...
+			cleanName = parts[len(parts)-1]
+		}
+		cleanName = strings.TrimSpace(strings.ToLower(cleanName))
+
+		foundApps, err := FindWorkflowAppByName(ctx, cleanName)
+		if err == nil && len(foundApps) > 0 {
+			foundApp = &foundApps[0]
+		}
+		if foundApp.ID == "" {
+			algoliaApp, err := HandleAlgoliaAppSearch(ctx, cleanName)
```
frikky, line 36895: Did we validate whether Algolia checks if it's cloud? 

Backend onprem doesn't have Algolia access

frikky (commented):
A few things:
- Timeout edgecases will be a nightmare if we enforce it
- The "Hybrid" also sounds like issues, as I don't see why it's necessary. I can't understand when or how the Agent itself is hybrid

Check comments (:

PR #479, opened 2026-09-16
fix: flaky test case for agent health check and a few other quirks with the agent

satti-hari-krishna-reddy:


ai.go
```diff
@@ -15994,26 +15980,38 @@ func GetOrgAiCredentials(ctx context.Context, callInfo AiCallInfo) (string, stri
 		for _, field := range auth.Fields {
 			// Check if the auth has a valid API key
 			if field.Key == "apikey" {
-				parsedKey := fmt.Sprintf("%s_%d_%s_%s", auth.OrgId, auth.Created, auth.Label, field.Key)
-				decrypted, err := HandleKeyDecryption([]byte(field.Value), parsedKey)
-				if err == nil {
-					curApiKey = string(decrypted)
+				if !auth.Encrypted {
```
frikky: It's probably better here to not fully trust auth.Encrypted, as this can also be controlled by the user (manual override)

You may want to just TRY decrypt, and if it fails, use the raw value.

ai.go
```diff
@@ -16029,6 +16027,17 @@ func GetOrgAiCredentials(ctx context.Context, callInfo AiCallInfo) (string, stri
 			foundModel = curModel
 		}
 
+		// Custom URL must only be used when paired with an API key
+		if len(curUrl) > 0 && len(curApiKey) == 0 {
```
frikky: Weird, wasn't this here already?

ai.go
```diff
@@ -16056,7 +16065,7 @@ func GetOrgAiCredentials(ctx context.Context, callInfo AiCallInfo) (string, stri
 		}
 
 		// Checks if cloud sync is set up
-		if len(org.SyncConfig.Apikey) > 0 {
+		if len(org.SyncConfig.Apikey) > 0 && org.SyncConfig.AiCloudSync {
```
frikky, line 16410: What is AiCloudSync used for?

frikky (commented):
:)

ai.go
satti-hari-krishna-reddy: AiCloudSync is a specific parameter introduced by Lalit that acts as a control flag. It requires the user to explicitly enable AI cloud sync. @LalitDeore , mind chiming in here with the exact details or context on this ?

LalitDeore: > What is AiCloudSync used for?

AiCloudSync is used to track whether user have enabled/disabled the ai execution through cloud sync or not. By default it is off and it can be enable from admin page.

frikky: Why would it be off by default? On by default obviously makes more sense lol

ai.go
```diff
@@ -11247,26 +11247,12 @@ func RunAiQuery(ctx context.Context, info AiCallInfo, systemMessage, userMessage
...
-
-	if len(aiApiVersion) == 0 {
-		aiApiVersion = os.Getenv("OPENAI_API_VERSION")
-	}
-
-	if len(orgId) == 0 {
-		orgId = os.Getenv("OPENAI_API_ORG")
-	}
+	// Fallback only. Using SHUFFLE_AI_* instead of AI_API_KEY so legacy env vars
+	// don't accidentally block Cloud Gemini. If this is set, it's intentional.
```
frikky, line 11609: This is a stupid rationale.

Fixing gemini implementation is the choice here (which I kin of did anyway) 🤦  The fallback and fallback is obviously better as it uses more standard API-keys and such. And most people won't use this shit anyway

frikky (commented):
Off by default seems wrong

ai.go
```diff
@@ -9907,7 +9917,7 @@ data_filter:
 
 	if project.Environment == "cloud" {
 		//completionRequest.Store = true
-		completionRequest.MaxCompletionTokens = 5000
+		completionRequest.MaxCompletionTokens = 16384
```
frikky: Why so specifically 2^14?

ai.go
```diff
@@ -11970,8 +11993,8 @@ func RunAiQuery(ctx context.Context, info AiCallInfo, systemMessage, userMessage
 		for {
 			iterations += 1
 
-			if iterations > 1000 {
-				log.Printf("[ERROR] Fatal - Too many iterations agent LLM stream. Breaking out of loop.")
+			if iterations > 10000 {
```
frikky: Wait, I'm confused. What are we doing here again? Why 10k?

health.go
```diff
@@ -4947,7 +4947,7 @@ func RunOpsAgent(apiKey string, orgId string, cloudRunUrl string) (AgentHealth,
 func getRealTempC() (int, error) {
```
frikky: Is this just for testing?

frikky (commented):
👍

PR #481, opened 2026-09-21
fix sso and user session issue

LalitDeore:
Issues Fixed:

1. Fixed the issue where the session token was being changed with API key.
2. Allowed users to access a different parent tenant if they already exist in that tenant for SSO.
3. Fixed the issue causing duplicate users to be created for the same email address.


frikky:
Looks alright

PR #482, opened 2026-09-21
added status>300 bad status notif

kshgrk:


shared.go
```diff
@@ -19762,15 +19762,16 @@ func ParsedExecutionResult(ctx context.Context, workflowExecution WorkflowExecut
 		}
 	}
 
-	if workflowExecution.Workflow.Configuration.SkipNotifications == false && actionResult.Status == "SUCCESS" && strings.Contains(actionResult.Result, "\"success\":") && strings.Contains(actionResult.Result, "\"status\":") {
+	if workflowExecution.Workflow.Configuration.SkipNotifications == false && actionResult.Status == "SUCCESS" && strings.Contains(actionResult.Result, "\"status\":") {
 		type resultMapping struct {
-			Success bool `json:"success"`
-			Status  int  `json:"status"`
+			Success *bool `json:"success"`
```
frikky, line 19767: Why would you want to make bool a pointer instead of pure bool?

shared.go
```diff
@@ -19762,15 +19762,16 @@ func ParsedExecutionResult(ctx context.Context, workflowExecution WorkflowExecut
 		}
 	}
 
-	if workflowExecution.Workflow.Configuration.SkipNotifications == false && actionResult.Status == "SUCCESS" && strings.Contains(actionResult.Result, "\"success\":") && strings.Contains(actionResult.Result, "\"status\":") {
+	if workflowExecution.Workflow.Configuration.SkipNotifications == false && actionResult.Status == "SUCCESS" && strings.Contains(actionResult.Result, "\"status\":") {
```
frikky, line 19765: What are some edgecases removing the "success" check can cause?

frikky (commented):
:)

shared.go
kshgrk: when you unmarshal json into a plain bool field and the key is absent, go doesnt leave it unset, it fills in the zero value, which for bool is false

example:

--- plain bool: can't tell 'absent' from 'false' ---
  input={"status": 404}                                           -> Success=false (always a real bool, never nil)
  input={"success": false, "status": 404}             -> Success=false (always a real bool, never nil)
  input={"success": true, "status": 404}              -> Success=true  (always a real bool, never nil)
  
--- *bool: absent key stays nil, present key points at true/false ---
  input={"status": 404}                                          -> Success=nil (key was absent)
  input={"success": false, "status": 404}             -> Success=-> false (key was present)
  input={"success": true, "status": 404}              -> Success=-> true (key was present)

shared.go
kshgrk: previously it only checked for success, which is happening now, but status standalone can also trigger

shared.go
frikky: Are there any dangers here if we handle it badly?

shared.go
frikky: Yes, that's what the code says. I'm asking: Why does that matter?

PR #484, opened 2026-09-23
Fix agent decision parsing when LLM output contains code blocks

kshgrk:
there were 2 bugs:
1. fence cutting of already valid input json yielding an invalid json
2. fence cutting on the nearest triple backtick because the nearest triple backtick was sometimes a fence inside a json string instead of end of the wrapper

fixes proposed:
1. dont do fence cutting on a json which is already valid
2. check after cutting if the json is valid or not, if not then move to the next backtick, if nothing yields a valid json then fallback to the nearest triple backtick


frikky:
oooh, this is scary. Do you have tests, including real scenarios, where this is necessary? I can see it's only a validator -> same if, if else, but it still is worrysome to just make changes to this one

FixContentOutput() is used a tooon

PR #485, opened 2026-09-23
[Fix] - entity is too big error issue in datastore

LalitDeore:


db-connector.go
```diff
@@ -1137,12 +1139,17 @@ func IncrementCacheDump(ctx context.Context, orgId, dataType string, amount ...i
 			if err := tx.Get(key, orgStatistics); err != nil {
 				if strings.Contains(fmt.Sprintf("%s", err), "no such entity") {
 					log.Printf("[DEBUG] Continuing by creating entity for org %s", orgId)
+				} else if strings.Contains(fmt.Sprintf("%s", err), "cannot load field") {
+					tx.Rollback()
+					schemaMismatch = true
+					errMsg = fmt.Sprintf("%s", err)
+					log.Printf("[ERROR] Schema mismatch on load for org %s (%s) - skipping write to avoid data loss", orgId, err)
```
frikky: If you roll back... won't it just do it again a few seconds later? This sounds like a proooooblem

frikky (commented):
Sounds like you are creating a recursion loop... somehow here

db-connector.go
LalitDeore: It should not stuck in recursion. We rollback and store increment in memcache and then try again. This happen every time when we get that error and this is how I wanted to do it.  Any suggestion how we can handle this differently? We cann't continue in this case as datastore drop whole daily_statistics array when two backend is not in sync and when one have updated struct in one backend and other one don't have it. I am gonna set alert for this log as well.

